PR #299, opened 2014-12-18
Graph based mapping of values

MaPePeR:
I wrote a Graph based approach to calculate the values for the Objects based on a Graph of `Conversion<T>`s

I kept the Type of the actual Objects generic, so we can write simple `String` Testcases and later use `SimpleStack`

Its by no way implemented into ProjectE, but i wrote some very little test cases to test that it can map Basic Conversions. 

I used 4 Different Types to set fixed values:
* `SuggestAndInherit` set a minimum value for the Object, that can later be overwritten by calculated Values
* `FixAndInherit` set the value. things that are created from this one us this value to calculate their own value. (Should be used for all the Default MC things)
* `FixAfterInherit` calculates the value for all things that are created from this the normal way, but set the value afterwards.
* `FixAndDoNotInherit` Set the value of this thing to a specific value, but disable it for conversions. (Use this and value = 0 to totally remove the value from it)

Missing: 
* Something to say that the value should be calculated normally, but not be inherited to dependent things.

This probably needs some testing, to check if its compatible with the way ProjectE sets the values at the moment. 

EDIT: maybe "propagate" is a better word than "inherit"?


MaPePeR:
i tested this with Minecraft now (See [this branch](https://github.com/MaPePeR/ProjectE/tree/testGraphMapperWithMC) for changed i made) and compared the default EMC value to the one assigned by the GraphMapper: [this Gist](https://gist.github.com/MaPePeR/b1074cbfcca0e1471032) contains the Differences:

| Item | Default | GraphMapper | Explaination|
| ------------- |-------------:| -----:|-------|
|minecraft:daylight_detector |771|783| Error in Default Mapping: 3*1(glass)+3*256(quartz) +3*4(slab) = 783|
minecraft:wool |48|0|Has a cycle-recype wool+bonemeal=wool, which is not yet supported by the GraphMapper
minecraft:hardened_clay  | 82 | 64| Default uses Philosopher Stone recipe instead of normal smelting => Default EMC Value too high
minecraft:fire_charge  | 362  | 330 | Default uses Recipe with Normal Coal instead of Charcoal => Default EMC Value too high
minecraft:water_bucket | 768 | 0 | i did not teach it Fluid recipes yet.
carpets, beds, paintings | | 0 | Side effect of wool having no EMC Value
All the ProjectE Items |  | 0 | that are crafted with Mobius Fuel, which has a A lot of Cycles with the philosohers stone and Fuel Blocks

ViolentNinjaD:
This seems like it could be a really useful system.

I recommend that you squash your commits though :smile:

MaPePeR:
yea, that would probably a wise idea. But for now i will leave them as is, so everyone can see my rambling ;)

MaPePeR:
So with the latest commits i brought the difference down to the missing fluid mappings/bucket thing, which is the problem of the [dirty code i threw into EMCMapper.java](https://github.com/MaPePeR/ProjectE/compare/testGraphMapperWithMC#diff-1) to test this and not a problem of the mapping itself.

```
          minecraft:hardened_clay 1 0: !        82 !=         64!
           minecraft:water_bucket 1 0: !       768 !=          0!
                 minecraft:potion 1 0: !         1 !=          0 
     ProjectE:item.pe_gem_armor_0 1 0: !  29551712 !=          0!
     ProjectE:item.pe_gem_armor_2 1 0: !  30258504 !=   30259272!
        ProjectE:item.pe_ignition 1 0: !    285776 !=     286544!
     ProjectE:item.pe_harvest_god 1 0: !    280800 !=     281568!
            ProjectE:item.pe_swrg 1 0: !    559360 !=     560128!
       ProjectE:item.pe_zero_ring 1 0: !    280646 !=     281414!
  ProjectE:item.pe_ring_iron_band 1 0: !      2112 !=       2880!
     ProjectE:item.pe_gem_armor_3 1 0: !  28709376 !=   28710912!
      ProjectE:item.pe_black_hole 1 0: !    280712 !=     281480!
 ProjectE:item.pe_evertide_amulet 1 0: !    417792 !=          0!
ProjectE:item.pe_volcanite_amulet 1 0: !    418176 !=     422784!
```

MaPePeR:
So i added the fluid recipes and the difference is now: 

```
         minecraft:lava 1 0:           0 !=         64!
minecraft:hardened_clay 1 0: !        82 !=         64!
```

`minecraft:lava` is now in there, because i used an `ItemStack`-version of the Fluid-`Block` to store the emc value for the fluid itself.

MaPePeR:
Code is now in #376

PR #328, opened 2014-12-24
Made EMC system use doubles

karelmikie3:
as the title states I made the entire EMC system use doubles.


LemADEC:
Did we check the performance impact of this change?

karelmikie3:
We haven't noticed any impact on performance and we are using it on a private server with a private modpack with around 200 mods.

LemADEC:
could we actually measure it with a profiler like Opis?

MaPePeR:
I'm more worried about accuracy than speed. So if we decide to do decimal digits I would prefer to see rationals.

Shalashalska:
Doubles have 15-17 digits of precision, so I really don't think that will be an issue. Or, you could just increase the value of everything by a factor of 100.

LemADEC:
there's no need for 15 digits precision, doubles takes more space in memory and more time to compute. So, what's the motivation for this change?

Shalashalska:
It means that items that have crafting recipes that split inputs (f.e. stone slabs, carpets, BC pipes) can have proper EMC values.

MaPePeR:
```
> 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1
0.9999999999999999
```

karelmikie3:
Sadly this pullrequest has been denied

PR #357, opened 2014-12-30
Red Matter Furnace: respect output sides

MaPePeR:
Before the RM Furnace always outputted items to the top of adjacent blocks. Now it uses the proper side, that is touching the RM Furnace to output.

View PR-diff with [?w=1](https://github.com/sinkillerj/ProjectE/pull/357/files?w=1)


williewillus:
I improved the sided interactions of the furnii in this (and tweaked the collectors a bit): https://github.com/sinkillerj/ProjectE/pull/485

Why don't you have write access to this repo yet lol

PR #362, opened 2015-01-01
Config Gui

ViolentNinjaD:
With this pull request, you can now launch Minecraft, click on the mods section, select ProjectE and click on the config option. This allows you to change config options without having to restart Minecraft.

I have also added some useful comments for developers which show more information about parameters and such.


sinkillerj:
I would like to keep this text only right now, and the comment modifications seam a bit redundant.

PR #364, opened 2015-01-01
Don't repair items while swinging them (Repair Talisman)

MaPePeR:
Some tools tend to abort the breaking of a block, when the item is repaired during the swing.


Kolatra:
I've got no objections. Nice little fix.

PR #390, opened 2015-01-14
Destruction Catalyst, Hyperkinetic Lens and Catalitic Lens Config Options

ViolentNinjaD:
This pull request allows you to configure the Destruction Catalyst, Hyperkinetic Lens and Catalitic Lens.

The only problem is that the "Item disabled for server admin." messagte on my config options is posted twice in the chat log. I couldn't find out how to fix it :disappointed: Hopefully someone else knows how to do this.

### DIFF
--- a/src/main/java/moze_intel/projecte/config/ProjectEConfig.java
+++ b/src/main/java/moze_intel/projecte/config/ProjectEConfig.java
@@ -28,6 +28,7 @@ public final class ProjectEConfig
 	public static boolean enableDarkFurnace;
 
 	public static boolean enableTimeWatch;
+	public static boolean enableDestructionCatalyst;
 	
 	public static void init(File configFile)
 	{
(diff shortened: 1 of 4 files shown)


MaPePeR:
probably need to check for `world.isRemote`

ViolentNinjaD:
Ok, can't change it for about another 16 hours :disappointed: I'll see what I can do then

MaPePeR:
No need for sad smileys :)

ViolentNinjaD:
There we go, I fixed the double message. I'll look into making the diff cleaner now.

sinkillerj:
This will be handled personally.

MaPePeR:
Since all these items should extend `ItemPE` there could be a helper method, that uses the unlocalized name to determine if the item is disabled or not.

sinkillerj:
@MaPePeR I had the same thought but sticking with the native Forge config system at some point it has to use a string as a variable name, something which I've had a bad experience with in the past with java. I'm open to your thoughts on it though.

PR #431, opened 2015-01-27
removed errorneous speed setting, made arrow home on nearest enemy

Heimdell:
Added tracing when homing.
The problem is in unability to set gravity for EntityArrow (it doesn't implement EntityThrowable).
But, when I set homing on enemy's hat, it started working. Also, could kill enemy beyound 1m obstacle now.

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -1,0 +1,1 @@
+export JAVA_HOME=/usr/java/jdk1.8.0_31/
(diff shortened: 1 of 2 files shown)


sinkillerj:
Pull manually merged due to unneeded .bashrc f0e7b297a4336270adfb6ff59471a965b5662803

PR #447, opened 2015-02-02
Improve javadocs.

jamierocks:


src/main/java/moze_intel/projecte/api/ITileEmc.java
```diff
@@ -2,36 +2,46 @@
...
 {
+
 	/**
 	 * Set the EMC value of the tile entity.<br>
 	 * Gets called only server-side. 
+	 *
+	 * @param value The EMC value
 	 */
-	public void setEmc(double value);
+	public void setEmc(int value);
```
williewillus: Umm what's the reason for this?
EMC is a double internally, and doing this will break binary compatibility (addons compiled earlier expect to see setEmc(double) and then crash when they don't)
I'm sure the mod itself would crash too because it uses this API call quite a bit

MaPePeR: > EMC is a double internally
Sadly this is not very consistent throughout the mod. :/

williewillus:
Request close: Does not do what PR title states - Destroys backward compatibility by changing and API method signature and only does one minor javadoc additions that probably isn't even needed (link to a vanilla class :p...)

jamierocks:
emc can't be doubles anyway

williewillus:
if you feel like it shouldn't be one, then write a big giant refactor that changes it all and submit it. Don't submit this critical change in a broken PR that doesn't compile and then call it "update javadocs"

PR #478, opened 2015-02-09
Add attribution

MidnightLightning:
Hey there; just noticed that this work had spun off of Equivalent Exchange. Just added in an attribution link for myself, if you don't mind. And if you need any further assistance with texturing, I'd be happy to assist!


sinkillerj:
I prefer to keep the formatting as it is now, but I'll toss in your URL. Apologies for not including a URL sooner, wasent sure what you would prefer. If you haven't already feel free to stop by our IRC chan.

PR #545, opened 2015-03-14
Make Collector&Relay tooltips read from Consants Class

Kolatra:
This would allow the values to be changed if ever wanted, and the tooltips will automatically update with the new numbers.


sinkillerj:
This would result in a incorrect value, I'm planning on working on a tick to sec method made primarily for the pedestal cooldown config, but which could apply to this too.

Kolatra:
How so? The values seemed correct when I was testing.

williewillus:
Superseded by #581

PR #588, opened 2015-03-29
Bring back the CustomEMC to the API

MaPePeR:
This should probably only be a temporary implementation until we figured out how we want to design the api in the future.

I could not really test it because all the Reika-Mods appear to try to set values for wrong itemids? Or that is a bug in the code that i have not yet found...


1Matthias:
Almost all of Reika's items are supposted to be unobtainable through project E unintentionaly. This is not a bug, and allowing Reika's items to be created will cause SEVERE issues in his mods. This is because Reika's mods are extremely finely balanced.

Kolatra:
Reika has talked to us about this, both on TeamSpeak on IRC, and I agree.

sinkillerj:
Using ProjectE is a choice, a choice by the user, not by Reika.

ReikaKalseki:
So is whether they harass me for getting RC/CC stuff too early and destroying their world. Spoiler: They do.

1Matthias:
If you mean that using ProjectE to circumvent Reika's techtree is acceptable, I have to disagree. If you meant something else, please ignore the next section. Certain sections of Chromaticraft, if you spawn in the materials before you have the progression/knowledge to use them properly, can destory your world. Also, there is an exquisite techtree in Reika's mods, which should not be skipped around in. How would you like it if another mod came along and promptly made your own high teir materials attainable from the equivilent of a single block of dirt/made your entire techtree worthless?

sinkillerj:
The only way to bypass progression would be enabling the tome which is disabled by default. 

As for harassment, your issues section is your issues section, if you do not wish to support other mods in support issues you should make your rules more clear on the subject.

ReikaKalseki:
I make my rules abundantly clear. The kind of players I am talking to heed them approximately never.

Kolatra:
Even with the tome disabled in the config, people will just cheat it in if they don't know about the config.

sinkillerj:
This conversation is over, I will not nerf ProjectE just because someone doesn't feel like moderating their own issues.

1Matthias:
How is it nerfing your own mod to help another mod author to be able to keep their own techtree intact?

MaPePeR:
Funny, how this pointless discussion arises in the PR, that kind of tries to allow reika to blacklist items again. 
But no one answered the question about wrong item ids that are used by reikas mods.

ReikaKalseki:
> But no one answered the question about wrong item ids that are used by reikas mods. 

???

MaPePeR:
When I logged the itemids that were fed into the api I could not find a single one, that was valid. 
Also I could not find any skipping of progression. 

Our do you call skipping 3 crafting steps a skip of progression?

sp64:
Don't bother working with reika, he's an entitled asshole who is not going to be happy with anything less than you bending over and letting him f*** you like the f** he is. I say this as another big developer (under a fresh acct so that he can't track me down and smear me). The proper approach to reika is pull all support (it would be nice if we could prevent him from being installed alongside, but we'd get in trouble for that), and ignore any requests or demands he makes. Nothing he says is worth listening to, end of story. I would also go around telling people about this convo so that if reika tries his DRM s*** again everyone knows about it and can uninstall his crappy mods.

sinkillerj:
@sp64 Lets try to keep things civil around here.

1Matthias:
It is actually not skipping three steps of crafting. The end tier of Rotarycraft takes about 80-100 dedicated hours of gameplay to acheve, with several gated steps of progression. For instance, to get tungsten, you need an extactor, which needs gasoline engines at minimum, which need a fermenter and a grinder, which need a) Dc engine+water+dirt+Blast furnace+Sugar, and b) Steam engines+fire+water. And did I even mention all of the resources that you need for those steps alone? All that I have mentioned requires about ten+ stacks of iron, not counting gold, diamonds, etc for shafts etc. Please use these mods BEFORE you make snap judgements.

MaPePeR:
You are funny. 
I DID try rotary craft with ProjectE and the most complex thing that had an EMC value was  gear boxes and hazard suits. 
And some stuff that has been ore dictionaried. (shards, hardened glass  and ingots) 

So:  PLEASE try ProjectE before you make snap judgements. 

And again, because you still have no clue how ProjectE works: even if all these things WOULD have emc(and they don't!!) you would still need to create at least one of everything to be able to transmute into that item. 


I'm really sick of this discussion. We should write code to give all the things emc to make the  discussion at least a little bit worth it. (that's not supposed to be a threat, but I'm REALLY sick of this.)

MaPePeR:
I think this PR suffers from itemIds changing between 'FMLInitializationEvent' and 'FMLServerStartingEvent'?

MaPePeR:
So i think this is kind of ready now. You now can control for each mod that uses the api for what mod it can set/remove EMC values.

sinkillerj:
@MaPePeR So far everything seams to be working well including the blacklisting, before publishing a new build I tested it against EREMC, may help with future tests: https://github.com/Ryokusitai/EREMC/blob/master/src/main/java/eremc/EmcHandler.java

http://forum.minecraftuser.jp/viewtopic.php?t=24695

MaPePeR:
Ah. That's indeed a good mod to test it with.

PR #592, opened 2015-03-31
Chisel2 Mapper [finished?]

MaPePeR:
Sadly its hard to test this without Chisel being present, because ForgeGradle adds it magically.

bdews implementation: https://github.com/bdew/ProjectE/blob/f1b08624ff47c6cc716576701024cdb38ff3d297/src/main/java/moze_intel/projecte/emc/ChiselMapper.java


MaPePeR:
Aaaaand it does not work if Chisel is not present, because to instantiate the class all the imports have to be successful. 

Anyone an idea how to solve this?

williewillus:
Don't add a mapper to the mapper list unless their requisite mod is present?

MaPePeR:
But i can't think of a way to determine the requisite mod for a mapper without loading the mapper

artman41:
I have a fix.

do the following:

image

MaPePeR:
I'm aware of this possibility, but it does not provide the flexibility i want.

MaPePeR:
With the latest commits the mapper now does not crash anymore when Chisel is not present.

I don't see any issues with this anymore. Please test and report.

PR #612, opened 2015-04-03
Allowed the Iron Band to be created with a Volcanite Amulet

Parker8283:
I was surprised that this wasn't in here already :P

Also cleaned up the Evertide Amulet's implementation of IFluidContainerItem; one less opcode :P

EDIT: Reverted implementing IFluidContainerItem on the Volcanite Amulet. Forgot that Lava was 64 EMC.

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/ObjHandler.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/ObjHandler.java
@@ -410,6 +410,7 @@ public class ObjHandler
 		
 		//Rings
 		GameRegistry.addRecipe(new ItemStack(ironBand), "III", "ILI", "III", 'I', Items.iron_ingot, 'L', Items.lava_bucket);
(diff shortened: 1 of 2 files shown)


williewillus:
The way it is currently the Volcanite costs EMC to create lava.
We have to retain that - the way you have it I could just stick it in a Fluid Transposer and get free lava

Parker8283:
Good point. Will fix momentarily.

Parker8283:
And by fix I meant revert :P

sinkillerj:
I wonder if there is a ore dict entry or something we could use to do this better.

Parker8283:
I wish, but I don't believe so. I checked how EE2 did it, and this is what they did.

MaPePeR:
Maybe we could do it similar to how gregtech damages tools when crafting and require the amulet to be charged somehow. 
That would complicate stuff, though.

Parker8283:
This particular recipe in EE2 was just a standard shaped crafting recipe, and didn't require any extra EMC to craft this. But obviously, we could change that if we wanted, but as you said, would be complicating things.

Issue #629, opened 2015-04-08
EMC Allocation Never Happens (Injection into Direwolf 1.7.10 pack)


abraxas:
I tried to plug ProjectE into the otherwise stock Direwolf20 pack seen here https://feedthebeast.atlassian.net/wiki/display/PML/Direwolf20+%281.7.x%29+1.2.1

Everything seemed to work in terms of recipes, but nothing has EMC values.  Not vanilla, not ProjectE stuff, nothing.

Here's my error log: http://paste.ubuntu.com/10772804/

It appears to me that EMC values are being allocated, then immediately stripped... All this, however, happens on the tail of a File I/O error that may or may not be related.

At the end of the process, I end up with an empty mapping.cfg.  I'm not sure if that's supposed to represent the EMC values for blocks.

For head's up.  I'm working on a Macbook pro on OS X Yosemite.  I have not yet tried this in linux or windows, but can update the ticket with results to those if that turns out to be material.

MaPePeR:
The problem is this:
```
Exception in thread "ProjectE Reload EMC Thread" java.util.ConcurrentModificationException
at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
at java.util.ArrayList$Itr.next(ArrayList.java:851)
at moze_intel.projecte.emc.mappers.CraftingMapper.addMappings(CraftingMapper.java:30)
at moze_intel.projecte.emc.EMCMapper.map(EMCMapper.java:31)
at moze_intel.projecte.emc.ThreadReloadEMCMap.run(ThreadReloadEMCMap.java:37)
```

line 16439

abraxas:
Is this sufficient for research on your part, or do you need any more from me?

Alternatively, does this scream of an obvious conflict that I could attempt to qualify/disqualify for you?

MaPePeR:
I think this is sufficient, thank you. 
Sadly i don't know about any way to figure out which other mod caused it. It is some kind of conflict, but its only conflicting because bad code on our side, i think.

sinkillerj:
It's def on our side, other mods are not to blame. A emergency build has been published which reverts the pull that caused this for now, however I will leave this issue open till a proper fix is in place.

Parker8283:
Never got this on my end. I'll do some testing and see if I can't fix it. Sorry to cause this.

Parker8283:
[...]

PR #632, opened 2015-04-09
"Fix" #629. Also log how long EMC (re)mapping takes.

Parker8283:
"Fix", as in, don't thread ServerStarting. Still think others shouldn't be registering recipes during that phase, but also, users should be able to have EMC values for their stuff. So, yea.

Also, at the request of @MaPePeR, the thread, along with ServerStarting, logs how long it took to (re)map the EMC values.


Parker8283:
uhh...what? git, you feeling ok?

Parker8283:
Let's try that again.

PR #633, opened 2015-04-09
"Fix" #629. Also log how long EMC (re)mapping takes.

Parker8283:
"Fix", as in, don't thread ServerStarting. Still think others shouldn't be registering recipes during that phase, but also, users should be able to have EMC values for their stuff. So, yea.

Also, at the request of @MaPePeR, the thread, along with ServerStarting, logs how long it took to (re)map the EMC values.

Take 2.

### DIFF
--- a/src/main/java/moze_intel/projecte/PECore.java
+++ b/src/main/java/moze_intel/projecte/PECore.java
@@ -136,7 +136,15 @@ public class PECore
 			new ThreadCheckUUID(true).start();
 		}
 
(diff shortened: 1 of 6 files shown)


Parker8283:
When IntelliJ's git client isn't working, CLI prevails!

sinkillerj:
Some conflicts were caused by the refactor, once those have been resolved in this pull I can merge it.

Parker8283:
Hang on while I fight with git

Parker8283:
Alrighty, ready to merge.

PR #636, opened 2015-04-10
Master

artman41:
- Added a new EMCMapper to allow for easier Manual Mod Integration.

- Added some key items from IC2 to the Mapper.

- Moved the one AE2 item to the mapper.

- fixed a typo in the console output.


artman41:
Hopefully this will allow for easier compatibility with mods and allow for integration with key mods that the current system doesn't support, such as IC2.

artman41:
Just noticed, the IC2 api contains multiple class files that weren't being shown in the package explorer in eclipse. Those files however should be easily deleted but, if need be, I can delete them and resubmit the pull request

artman41:
Noticed that the IC2 Api was packaged with obfuscated code, deleted that to just leave behind the java files which are needed.

sinkillerj:
We do not manually define EMC values for mod items, nor do we plan to.

PR #649, opened 2015-04-13
Pregenerate emc

MaPePeR:


MaPePeR:
This has to be changed, so that it does not throw an error, when an item is not found, so mods can be removed without destroying the pregenerated values.

Issue #682, opened 2015-04-30
Config Option to get the old right-click ability for RM/DM Pickaxe


MaPePeR:
Before the refactor the DM and RM Pickaxe did magically suck nearby ores from the ground when activating its right-click ability. (Instead of the vein-mine when right clicking oreveins)

I think people liked that ability a lot and now want it back.

PR #683, opened 2015-04-30
Vein Mine with DM/RM Pick like before.

MaPePeR:
Fix for #682.

Tested in Singleplayer


williewillus:
Was this how it was in EE2?

MaPePeR:
I have no clue. But i liked it how it was before.

And it made the charges worth it. Increasing the range for vein mining ores feels very useless because ore veins are very small by default.

MaPePeR:
Nevermind the latest commit message. i had problems wrapping my head around `!isRemote` vs `isRemote`. It should be all fine. (It even worked in my test! - Singleplayer only though)

MaPePeR:
@williewillus is there a reason your vein mine code does not use `ForgeHooks.canToolHarvestBlock(block, world.getBlockMetadata(i, j, k), stack)`? I remember that being needed to make it work with Tinkers Construct Ores. A
And i just had reports that these don't work anymore with the new vein mining functionality.

williewillus:
@MaPePeR Not my veinmining code, I just extracted it from somewhere else and fixed a bug :p feel free to change it

src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
```diff
@@ -566,4 +568,39 @@ protected void tryVeinMine(ItemStack stack, EntityPlayer player, MovingObjectPos
...
+	 * Based on https://github.com/sinkillerj/ProjectE/blob/5f3734769483ee00fa6dce2fa2c96c4b18aaa218/src/main/java/moze_intel/projecte/gameObjs/items/tools/DarkPickaxe.java#L149-L172
+	 */
+	protected void mineOreVeinsInAOE(ItemStack stack, EntityPlayer player) {
+		if (player.worldObj.isRemote)
+		{
+			return;
+		}
+		int offset = this.getCharge(stack) + 3;
+		AxisAlignedBB box = player.boundingBox.expand(offset, offset, offset);
+		//CoordinateBox box = new CoordinateBox(player.posX - offset, player.posY - offset, player.posZ - offset, player.posX + offset, player.posY + offset, player.posZ + offset);
```
williewillus: CoordBox is dead :p nuke this line

williewillus:
@MaPePeR also, you could just change PEToolBase's canHarvest method to query forgehooks as well

src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
MaPePeR: Yea, I left it as a comment to let someone else check if these lines produce the same box. :)

williewillus: but....it's dead code...few will care about the source of it (like that big ugly long github link right there :p)....better to put these notes in the PR or the commit rather than muddy up the codebase with them

MaPePeR:
@williewillus that does not work because the canHarvest Method does not allow metadata.

PR #691, opened 2015-05-04
NEI Mapping

MaPePeR:
It already kind of works! (Only tested in singleplayer, though)

But it requires you to run `/projecte_reloadEMC` after loading the world and then use the pregeneration-option, after you have finished changing values/configs.

(Remember: the `/projecte_reloadEMC` command actually rereads the `mapping.cfg`, so no need to restart the game)

Example i used to test it: EnderIO Sag Mill

```
            crazypants_enderio_nei_sagmillrecipehandler {
                # Has 31 recipes for "SAG Mill" with identifier "EnderIOSagMill" [default: false]
                B:enabled=true

                # Enable extensive logging for this RecipeHandler [default: false]
                B:logging=false

                # Increase the amount gained. [range: 1 ~ 128, default: 1]
                I:outputMultiplier=1
                I:skipSlots <
                    2
                 >
            }
```
`logging` prints all the Recipes from that RecipeHandler to the log, to figure out the slots needed for the `skipSlot`-Option. 

The `skipSlot` in the example makes it not read the 'flint/dark steel ball'-slot of the sag mill as an ingredient.

The outputMultiplier is intended for the Potions. The NEI GUI only shows 1 potion as output, but as we all know you can create them in batches of 3.

For now this is not really intended for merging, but as a reminder that it exists.


MaPePeR:
Has some problems with AE2's Shaped Crafting Recipes. Especially the ones to color cables:
Because the displayed itemstacks have a stacksize of 0 it does not calculate the right costs for the colored cables.

williewillus:
Wait does this scrape nei plugins to map EMC values? :P

MaPePeR:
Exactly.

sinkillerj:
The plan is to implement this after 1.7.1, I may make a few minor usage tweaks of my own.

MaPePeR:
I think i will extract this functionality into an extra mod that provides commands to generate a json file for the recipes.
Then have a Mapper that reads such "custom conversion" files. We could even setup a website to share these files for different mods, so you can pick what you want.

That should be much more flexible and easy.

MaPePeR:
@sinkillerj I'm not sure if an extra mod is a good idea. at the moment i have commands planned like this:
```
/neircollect collect <classname with auto completion> <some options to change stacksizes or ignore slots and logging>
/neircollect store <filename> 
```

i think it would be much easier to just add it to ProjectE, so i can use the json-classes from #860.

Yay/Nay?

PR #711, opened 2015-05-11
Don't clear emc map until mapping is finished.

MaPePeR:
As of now this is not tested at all.


MaPePeR:
Not needed anymore because thread has been removed

Issue #718, opened 2015-05-15
Typo in Relay MK2


Zoryn4163:
The internal/unlocalized/whichever (not overly familiar with terms) has a typo. 'Realy' instead of 'Relay'
See: ![img](http://i.gyazo.com/6777bbf66ec1bf58cfc444cd82ed2ad4.png)
Using: ProjectE-1.7.10-PE1.4.6

Kolatra:
Okay, I am going to fix this, and the rest of the UNs that have spaces in them.

sinkillerj:
Of all the possible typos that one has to be the funniest. 

![](http://www.perivision.net/wordpress/wp-content/uploads/2011/11/rageface-really.jpg)

ErrorCraftLP:
Fixed in #720.

PR #719, opened 2015-05-15
Fix Internal Names

Kolatra:
This removes the spaces from them, and fixes the Relay MK2 typo that has been around probably since they were made. This will break the blocks in saves, but in the long run, this will be better, we could also use that remap method in PECore to work around this like we did with items.

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/ObjHandler.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/ObjHandler.java
@@ -227,26 +227,26 @@ public class ObjHandler
 	public static void register()
 	{
 		//Blocks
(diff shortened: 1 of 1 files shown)


williewillus:
Save compatibility, especially within the same MC version, should always be THE number one mod priority.
I don't support merging this at all until we work out a way to convert saves (FML remap event?)

sinkillerj:
I agree with @williewillus , you should be able to follow the remap code the is already in place as a basic template, so most of the work is already done for you. 

Also tagging: #718

Kolatra:
It has been done, the remap code in the Core Class is a great way to do it, hopefully we can do this, because honestly these should have been changed a LONG time ago.

williewillus:
Yeah I'm looking at it rn and it doesn't look bad at all.

Kolatra:
Extended in #720.

PR #725, opened 2015-05-17
Prefix-based Tab completion and a map for the commands.

MaPePeR:
If you write `/projecte se<TAB>`  it will suggest `/projecte setEMC` instead of `/projecte changelog`


MaPePeR:
@sinkillerj please consider merging this as well. The broken tab completion is annoying :/

PR #764, opened 2015-05-31
Lootball Despawn and Pedestal Timewatch Fix

abaw7:
Noticed that the Lootball's despawn time is set to 6000, not sure why, changed it to 300 like every other item in the game. Also noticed that the TimeWatch still worked in a pedestal if it was disabled. Added checks to prevent it and notify the player that the item is disabled.


williewillus:
Um, if I'm not deathly mistaken, 6000 is in ticks. Like every other number in the code. 6000 ticks @ 20 TPS = 300 seconds.

williewillus:
Also, I do not agree with the addition of a new config. The Pedestal configs were designed in a way to be non-redundant - You don't need an enable=true setting and then set a number for the cooldowns and bonuses when you could effectively disable the timewatch effect by just setting both timePedBonus to 0 and timePedMobSlowdown to 1.0.

MaPePeR:
@williewillus i think abaw7 did not add a new config, but used the 'enabled Items in ProjectE'-config to also disable its effect in the Pedestal. (See, there is no change to the PEConfig-file)

abaw7:
That's exactly what I did there. And, every test I did (Made sure to log the TPS at 20 the whole time) it was measured in seconds. I could be wrong on this, but that's what every test of mine showed.

williewillus:
Perhaps we should change those configs to read "disable recipe" because that's all they do for the moment. They're a long shot from actually disabling things.

williewillus:
And that's really strange for the timing. I'll recheck

abaw7:
Actually, running a server, I can say the only thing they disable is use of the item. Players are still able to craft it. The OBJManager doesn't have it being removed due to it being a crafting ingredient in something else.

williewillus:
Heh, TIL.
Perhaps we should do what ic2 does and quit registering the block completely when it's disabled amirite /s

abaw7:
That's what I've been patching it to do since I didn't feel like following the source code to disable it to the standard set by the config option. But, that raises the problems with the Gem Legs Crafting recipe. since that requires the Watch

abaw7:
Just as proof. This is using the current code base + My pull request. http://imgur.com/N3sIbIy

williewillus:
Oh, no. Removing blocks when they're disabled is a terrible idea. Breaks save compatability. I was being sarcastic

williewillus:
Also, I just tested myself setting the lifespan to 300 and the ball despawned around 17 seconds after it was created....

abaw7:
Alright, I'll revert that change in my code. The pedestal change should be good though right?

abaw7:
@williewillus Just bumping to make sure it's not forgotten. I reverted my change to the Despawn timer, so the only change is with pedestals and the TimeWatch. (Pedestals with a TimeWatch in them won't work if the Watches are disabled by the current TimeWatch active option. They still work with everything else.)

PR #791, opened 2015-06-15
Update zh_CN.lang

Joccob:


autopsychorhythmia:
item.pe_zero_ring.name=零度指环
    pe.arcana.mode.0=零度
    pe.key.armor_toggle=护甲功能
    pe.key.charge=充能
    pe.key.extra_function=其他功能
    pe.key.fire_projectile=发射宝珠
    pe.key.mode=转换模式

plz.

Joccob:
The translation of 寒冰指环 is not my idea, I just copy from the last translator, but I change it anyway. As for "pe.arcana.inactive=关闭!", the " ! " is from the en_US.lang, I think I will keep it that way. And thank you for pointing out my inappropriate translation.

autopsychorhythmia:
thanks~ this translation of zero ring is same as its name in EE2.

PR #808, opened 2015-06-26
Adds clay to tools

Lilylicious:
Simple PR, makes it so that the morningstar and shovel consider clay a valid material to harvest.

I don't have a local copy of the repository, so this is untested. I am also not sure this is all that needs to be done to ensure that they harvest clay properly. The goal was to make the morningstar harvest clay on its 9x9 mode.

In most situations this is slightly convenient but not very important, but when using Pam's Clay where clay worldgens.. everywhere.. well..

Let's just say my enormous underground caverns has a ton of clay floating around, as the morningstar ignores it.


Lilylicious:
Rebased in some changes to RedStar.java, I believe that should make clay behave like gravel on right-click.

Lilylicious:
Fucked up something, don't mind that.

Lilylicious:
Actually tested now, works as intended. Is picked up with a general AOE harvest as well, and is vein mined just like gravel when focused on.

PR #884, opened 2015-08-01
[DONE] Prevent "Learned" and "Unlearned"-Text from overlapping

MaPePeR:


sinkillerj:
I was originally going to do this when I made the PR in the first place, the problem is it causes learned to not show up until the unlearned timer is over and vise versa, given this issue I'm not keen on merging this.

MaPePeR:
hmm? i don't think so... when i tested it it overwrote the unlearned text right away...

MaPePeR:
I just tested again and it worked properly. Maybe you did set the counter at a different place in the code?

MaPePeR:
@sinkillerj Maybe this can also go into the upcoming bugfix-release?

Issue #885, opened 2015-08-01
Repair talisman + Chisel2


Tencao:
Repair talisman in your inventory while using the chisel from chisel2 causes items to go missing and item duping

Put it here on will's request

MaPePeR:
Can be closed now, i guess.

PR #888, opened 2015-08-04
Repair talisman

### LINKED ISSUE #878: [Request] Use of Talisman Repairing of items in the Bauble slot.
Would it be possible for the talisman of repair to repair items in the bauble slot, such the Talisman of Remedium from Thaumic Tinkerer, that only get repaired when outside of the baubles slot. Sorry if this is oddly worded. o/

Lilylicious:
Closes #878 and #885. 

#885 is suboptimally fixed, but it's a step in the right direction (prevents items going missing and reported duping, since it doesn't repair while it contains items). Optimally we'd repair it when those things would not occur, instead of not repairing at all.

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/items/RepairTalisman.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/items/RepairTalisman.java
@@ -2,6 +2,7 @@ package moze_intel.projecte.gameObjs.items;
 
 import baubles.api.BaubleType;
 import baubles.api.IBauble;
+import com.cricketcraft.chisel.api.IChiselItem;
 import com.google.common.collect.Lists;
 import moze_intel.projecte.api.item.IAlchBagItem;
 import moze_intel.projecte.api.item.IAlchChestItem;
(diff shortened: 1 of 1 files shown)


MaPePeR:
You changed indentation in a weird way.

`bInvStack.getItem() instanceof IChiselItem` is a very very specific fix... i don't know why it dupes items when it is repaired, though....

Lilylicious:
I'll take a look at my settings later, didn't notice any indentation
weirdness.

I figured a very specific fix followed by a general fix was better than
waiting before a general fix. Was doing baubles anyway.

williewillus:
These both crash at runtime when the respective mods are not present.

Lilylicious:
.. Ah. Wait until I fix that before merge.

williewillus:
Move the chisel check or bauble repair to a separate @Optional.Method method, and call that method only when Loader.isModLoaded("modname") passes.
You might not need @Optional.Method, but it's a good safeguard to have

Lilylicious:
Will do, oversight on my part. Thank you for noticing.

Lilylicious:
All fixed, tested without mods and it works fine.

PR #890, opened 2015-08-04
Condenser display

### LINKED ISSUE #790: Feature Request - Stored EMC in condensors
As players have constantly asked me about this, is there a way to add a way to show the stored emc inside a condensor when there is no target item, and if possible, is there a way to make the bar show and collect emc when there is no target item, and maybe store up to x emc, maybe 500 mil emc.

Lilylicious:
Look at https://github.com/sinkillerj/ProjectE/commit/3672ba050b6a55b202a4505f44a0a0a58730680a to see the actual changes made, formatting obscured everything.

Closes #790, makes the whole process so much clearer. Especially when converting low EMC things into high EMC things in the MK1 condenser, as you can see from the screenshots in the following album.

http://imgur.com/a/Yv0Zq

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/gui/GUICondenser.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/gui/GUICondenser.java
@@ -40,7 +40,7 @@ public class GUICondenser extends GuiContainer
 	@Override
 	protected void drawGuiContainerForegroundLayer(int var1, int var2) 
 	{
-		int toDisplay = tile.displayEmc > tile.requiredEmc ? tile.requiredEmc : tile.displayEmc;
-		this.fontRendererObj.drawString(Integer.toString(toDisplay), 140, 10, 4210752);
+		String displayString = Integer.toString(tile.displayEmc) + "/" + Integer.toString(tile.requiredEmc);
+		this.fontRendererObj.drawString(displayString, 140, 10, 4210752);
 	}
 }
--- a/src/main/java/moze_intel/projecte/gameObjs/gui/GUICondenserMK2.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/gui/GUICondenserMK2.java
@@ -40,7 +40,7 @@ public class GUICondenserMK2 extends GuiContainer
 	@Override
 	protected void drawGuiContainerForegroundLayer(int var1, int var2)
 	{
-		int toDisplay = tile.displayEmc > tile.requiredEmc ? tile.requiredEmc : tile.displayEmc;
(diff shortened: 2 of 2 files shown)


sinkillerj:
I'm still highly against this, it makes it look tacky and defeats the purpose of the bar.

Lilylicious:
It's up to you if it's merged in or not.

Arguments for:
It's frequently requested. 
It clarifies what's going on. I personally got confused for a while when I was pumping in massive amounts of cobblestone (EMC 1) into a condenser and it didn't stop when I stopped inputting items (it absorbed faster than it produced).

The bar only helps when you want to see how close you are to the goal (current < goal), doesn't help when you have a massive buffer and want to know how much you have in there (current massively> goal)

Carbon-Ghost:
The best thing would be to add the current EMC progress in the bar when you hover your mouse on it ^^ If that wasn't already done, I don't remember, for me Project E is endgame because I have heavily nerfed recipes, isn't there yet, I think I might need to test in a creative world ^^

Lilylicious:
That's a good suggestion, a compromise between the two positions. Emphasizes the bar more, but keeps the info there if you look for it. 

Personally I still like the way it looks in the album I linked in the first post, but that's just me. If you prefer  @sinkillerj, we could move it into a tooltip hover-over for the bar or scrap it entirely.

Tencao:
This has been a longstanding request by players, so some solution would be greatly appreciated

Aiur850:
I also would like to be able to tell how much EMC is buffered in a condenser.

MaPePeR:
Showing the EMC value of the target item is totally useless, because that can be easily seen by hovering it and it does not change.

williewillus:
It's just inconsistent right now. Given that your target is diamond (8192), the bar at the top shows "Overall contained EMC / 8192" which means it stays at full after the tile goes over. The numbers too, it stays at 8192 as well once you go over. 

The point was we just need a way to see the condenser's current emc instead of a full bar and the number of the target just sitting there while it spits out more stuff without end

williewillus:
are we going to see an ACTUAL solution to the problem I described above? (doesn't have to be this PR, it's just a problem that needs a solution)

sinkillerj:
The GUI works as intended, the bar is progress towards the completion of an item, a full bar indicates items are being produced, there is no reason to show the EMC total.

williewillus:
I count at least 6 people from this thread (so, all -1 of them) that want to see the total, which means probably many more players want to as well

Didn't know we turned into Botania all of a sudden :P

sinkillerj:
Ill compromise if its done in a way like Carbon suggested.

williewillus:
I'd be fine with that

PR #892, opened 2015-08-04
Grand Alchemical Guidebook

Lilylicious:
Here's the collected work of me and Willie for the past month.

Important note regarding pre-existing files:
* ObjHandler reworked the way it registers items. Functionally the same with a few added extras related to the guidebook. 

Regarding the ManualFontRenderer, it's lengthy and a mess, but the most elegant way short of writing a whole new fontrenderer to adjust the private function I needed to. Would've been a cakewalk if it had been public, but nope.


sinkillerj:
Overall I'm ready to merge and work on my graphical tweets while Willie works on content, I do have one simple concern though regarding the objhandler changes. 

1. Is there any reason why its a generic "blocks" and "items" instead of PEblocks and PEitems, this would be safer and saner.
2. Could you briefly explain why the objhandler changes are needed.

Lilylicious:
1. For me it's because that's the way it used to be. @williewillus?
2. May be completely redundant at this point, after a whole bunch of
refactoring. Again, @williewillus?

williewillus:
Oh, they were there from my changes? I'm pretty sure I removed it if it was me .-.
It was in there for a brief moment when I was toying around with automatically registering everything by grabbing it off objhandler. But somewhere along the line we went back to manually declaring everything in ManualPageHandler. So it can be removed, I guess, if nothing else uses it

Lilylicious:
Is there any harm in registering in a function? If not, can just remove the
part of the function that does the .add thing and see if anything breaks.
If not, great.

williewillus:
You can revert the entire objhandler (except the book ofc) back to how it is on master rn. Sorry I left something so big behind uncleaned :P

Lilylicious:
I'll do it later today if sin doesn't beat me to it.

Lilylicious:
Objhandler is now the same as master, except for added manual related lines (item variable, gameregistry for it and recipe). I formatted the file to conform to the standards at the same time (was indented with tabs, not spaces)

sinkillerj:
If the file was using spaces I'm not sure why, we use tabs only.

Lilylicious:
willie is a clueless liar then! I'll do another commit shortly going back to tabs.

williewillus:
Sorry >.<

Issue #474, opened 2015-02-08
QED bugging on transmutation table


Zaflis:
QED is another block from Extra Utilities. It has a nasty habit of changing its item name constantly. Such as Quantum Enrichment Device, Quark/Electron Director etc...

This causes a big fuss on the transmutation table, such that trying to take one out puts it right back and refunds the EMC value. Server and client aren't in sync in this either, so if i search for "tra" for "trash can", it might pick up QED when it happens to go that slot. Just that my client doesn't see QED on the table itself. So it's actually changing its search name with that, and happens to include "tra" in its modname! exTRA utilities. I can often times pick items out from the empty slots too, so definitely something is not in sync with what is on table actually.

ProjectE-1.7.10-PE1.1.0

MaPePeR:
This can be closed now... hopefully... finally :heart:

PR #901, opened 2015-08-09
[DONE but risky] Send search result instead search string in transmutation gui

### LINKED ISSUE #265: QED derps
cant take qed out of transmutation tablet,  and if shift clicked it comes out but emc is restored to tablet and QED goes missing dev 10 version

MaPePeR:
This should finally fix #265 and #474 and also allows to search for localized item names on servers.

Needs testing on a server...

How it works:
When both server & client call `TransmutationInventory.updateOutputs()` the Client will execute the method before the server and will send its ItemStacks in the output slots to the server. (This is caused by inventory clicks being evaluated on the client before the WindowClickPacket is send to the server)

When the client wants to update the outputs based on an Event that is not fired on the Server (keyPress, mouseClick into Searchbar) it uses `TransmutationInventory.updateOutputs(true)` which will result in a Packet that the server will write into its output slot without waiting for a call to `updateOutputs`.

This is needed, because otherwise the server would write the new output stacks into its internal slots before it recieves the slotClick-Event.

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/container/inventory/TransmutationInventory.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/container/inventory/TransmutationInventory.java
@@ -313,13 +313,13 @@ public class TransmutationInventory implements IInventory
 
 	private void readUpdate()
 	{
-		try
+		List<ItemStack> newOutputSlots = serverOutputSlotUpdates.poll();
+		if (newOutputSlots == null) {
+			throw new RuntimeException("Server could not read output-slot-update from client. Playername: " + this.player.getCommandSenderName());
+		}
+		else
 		{
-			List<ItemStack> newOutputSlots = serverOutputSlotUpdates.take();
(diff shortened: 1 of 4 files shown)


sinkillerj:
This could use thorough testing, I may need to put out a alpha for this one. Regardless from a code standpoint it looks pretty solid.

MaPePeR:
With this we could bring back the NEI enhanced filtering from #377, btw.

sinkillerj:
I have no plans to ever bring back NEI search bar integration.

MaPePeR:
Why not? It allows for better searching. And it only caused issues because the search happened server side as well. 

Also I need to sneak a check in there, that checks for the player knowledge. At the moment a bad client could fake knowledge, I think.

MaPePeR:
#921 has this PR extended with the NEI Enhanced Searching. I would prefer to test both of these at once in an alpha version

sinkillerj:
Do both need to be merged or just one? If just one close the dupe and resolve merge conflicts please.

PR #918, opened 2015-08-14
[Done] Fix Mk2 Condenser not stopping to consume items

### LINKED ISSUE #914: Condensor MK2 does not have an upper limit on internal emc storage
so it will always consume the items and never 'fill up' - even if the output inventory is full.

I wonder what happens if that internal emc storages overflows or becomes Double.INFINITY...

MaPePeR:
fixes #914


blackdragonIV:
can you still condense items with crazy emc number ex. 8x compressed cobblestone from extra util

williewillus:
Nothing has changed, this just makes the condenser mk2 stop absorbing items when the output side is full

PR #921, opened 2015-08-16
[DONE but risky] NEI enhanced searching in transmutation gui

MaPePeR:
Based on #901, but since we need to do an alpha for that anyway, we should also include this PR in the alpha, because it might break stuff, too. 

Use NEI's Advanced Filtering Mechanics in the Transmutation GUIs
    
Allows to search for itemIDs, metadata, modnames and all the other cool stuff the NEI search bar has to offer.

Tested in Singleplayer and Server with NEI. Did not do a no-NEI test, yet, though.


MaPePeR:
@sinkillerj fixed you some merge conflicts.

Issue #353, opened 2014-12-30
Philosopher Stone Weirdness


Shad0wB1ade:
Tested with both dev 11 and 12 on forge 1277
It basically does the block next to the block your pointing at while pressing shift and not looking at the center
See the quick youtube video:
https://www.youtube.com/watch?v=HfprUN9CHCk

MaPePeR:
Actually this IS a vanilla bug: the Philosophers Stone uses the `getMovingObjectPositionFromPlayer`-Method to find the targeted block. (This Method is also used in CataliticLens,DestructionCatalyst,DiviningRod*,MercurialEye and i could recreate this with the destruction catalyst as well).
In Vanilla this method is used by the Bucket: 
Create a 1x1 Hole and place a bucket of water in it. Now go to the very border of the water source block and right click it with a bucket: you will see the watersource disappearing for a very short moment, but instantly reappearing, without filling your bucket.

Shad0wB1ade:
So I was actually right? YEY I apparently know what I am talking about...

Kolatra:
Fixed by https://github.com/sinkillerj/ProjectE/commit/7632d8c8ebb93dba917029fff4cfed8627f808b7.

PR #925, opened 2015-08-16
[DONE] Fix Lava -> Obsidian Transmutation with Philosopher stone, fixes #924

### LINKED ISSUE #924: Cannot transmute Lava to obsidian using philosopher stone
Probably caused by [this](https://github.com/sinkillerj/ProjectE/commit/5a9165f7cdf839a48def12fb2c9c654204b0f3a0) commit from ages ago, that fixed another issue.

`#minecraftforge` claimed that the bug in `getMovingObjectPositionFromPlayer` was fixed in 1.8.

@williewillus could you take a look at that method in 1.8, so we maybe can get that fix into 1.7?

MaPePeR:
Revert "Fix Philosophers stone sneak right click."

This reverts commit 5a9165f7cdf839a48def12fb2c9c654204b0f3a0.

This reopens #353, though.. not being able to transmute the lava is a bigger bug, i think...

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/items/PhilosophersStone.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/items/PhilosophersStone.java
@@ -28,6 +28,7 @@ import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.Direction;
 import net.minecraft.util.MathHelper;
+import net.minecraft.util.MovingObjectPosition;
 import net.minecraft.util.StatCollector;
 import net.minecraft.world.World;
 import net.minecraftforge.common.util.ForgeDirection;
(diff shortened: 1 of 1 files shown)


williewillus:
I just looked at Item.getMovingObjectPositionFromPlayer 1.7 vs 1.8 and I don't think fixable in a small amount of lines.
It got fixed in 1.8 alongside another big refactor in the MC codebase (entities being referenced by eye height on client vs feet origin on server)

But I think if it's to fix broke transmutations I would pull this and live with the "you get the wrong block when sneaking" which is just a vanilla 1.7 bug.

MaPePeR:
Ok.Thanks for looking at it. I had the hope it would be a small `if isSneaking`, but if that is not the case it is not really important to fix this little corner case.

sinkillerj:
Why would you ever even need lava to obsidian? #OvercomplicatedBucket

williewillus:
It was one of the original ee2 transmutes iirc

sinkillerj:
Yea but does anyone have objections to its removal? It seams like the sane solution here.

williewillus:
I don't really care, as long as my precious obsidian->lava isn't removed :D
Also while we're at it, water->ice can go too.

sinkillerj:
Leaving obsidian to lava would create a one way world transmutation.

Carbon-Ghost:
But what about the freezing Zero ring ? You can turn lava into obsi and water to ice IIRC, so why remove this and keep this one ?

sinkillerj:
Thats the specific job of that item, that would not be removed. In fact its even more reason to change this, the Philo Stone makes those features pointless.

MaPePeR:
We either should remove the Recipe from NEI then - or fix it.

williewillus:
But muh exploit :(
I vote to keep the transmutations in because I'm sure some player out there will get upset if it gets removed (and probs way more upset than I am lol)
The compromise is to just let a vanilla bug back in that is an edge case and will be fixed in 1.8 already

williewillus:
And if we're talking about duplicated features half the items duplicate features from some other item lol

MaPePeR:
I'd rather have the functionality than the right click fix.

Carbon-Ghost:
Well if it's a serious bug ( I still am lost and don't understand what is the bug ) then fix it. 

Wait I think I get it, it makes flowing lava into obsidian and then obsidian into lava ? Then it's here since 1.2.5 as it's the same for water ( less important since it's infinite ).

williewillus:
The bug is very minor, it's if you're looking at the very edge of a block and then shift so your crosshair is over another block, it'll still act on the original block. It's a vanilla bucket bug that was fixed in 1.8.

MaPePeR:
So #1056 is the alternative fix for this, then...

williewillus:
1.7 no longer supported so this is irrelevant 
*shouldn't* be a problem now

PR #931, opened 2015-08-16
Add new line to applicable files to sanitize project

HashBang173:


williewillus:
No.

For one: Read number 4 of this: http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/
For two: I bet these are not the only places this happens, it's going to take a lot more to change it all
For three: This doesn't sanitize the project, this pollutes our commit history with useless reformats :/
For four: Unless we all use IDEA with the same checkstyle configs or whatnot, this isn't gonna last very long :p

sinkillerj:
All cleanup and refactoring is under the authority of williewillus, you may certainly suggest changes to him, but as it stands this pull is unneeded.

PR #945, opened 2015-08-19
Fix for Life Stone infinite EMC consumption

TheKysek:
After merging https://github.com/sinkillerj/ProjectE/pull/852 Life Stone wasn't stopping to consume EMC even if player had full hunger and health.
This change I made seems to fix it.


MaPePeR:
This line just makes sure that the item always has 128 EMC stored or it will turn off automatically... i don't see this causing infinite emc consumption, because the stored emc is then only used from the item, when there is health or hunger to restore...

TheKysek:
I also have no idea why the change you made started causing problems with Life Stone nor why the change I made fixes it. I think that that problem should be looked into by someone more familiar with the cadebase than me.

MaPePeR:
#946  is the proper fix.. changing that if apparently triggered a bug in `consumeFuel`... very weird.

TheKysek:
Thanks for fixing it!

PR #949, opened 2015-08-19
Custom recipes

Lilylicious:
NOTE: This is ready to be merged as it is, but there's one thing that I want input on before I put any work into it. The custom recipe RecipesCovalenceRepair. It's definitely something I want to get into NEI, but someone thought that @sinkillerj might've mentioned that he wants to do some changes in that area. 

The question is: Are the covalence recipes going to change in the near future, and should I thus avoid working on them until such time? If they won't change, I have some ideas on how I can represent it in NEI. I think I would do a recipe for each covalence dust count, with the tool being cycled between the applicable ones, and the output matching the current tool. Find usages on covalence dust would then produce a tab in NEI that shows each dust count as well as what tools are repaired with how much dust. The other option is a text screen explaining that you can repair with dust.

---

Did some changing in two of the custom recipes, AlchemyBag and KleinStar. Instead of having a single instance of the recipes that has a catchall matches() method, they're now properly divided up into multiple instances (30 for the alchemy bags, 5 for the klein stars).

This meant adding constructors and changing how the matching() method worked. When registering the recipes, they're now added with output/input (for klein stars) and output/inputBag/inputDye(for alch bags). This simplifies the matches() method as well as conforms to how recipes usually work in Minecraft code (looks similar to Shapeless/Shaped recipes now, which they are).

The reason we need the custom recipes is to retain NBT data through crafting (klein stars with charge upgrade into the next step with the same charge, etc).

The alchemy bags are added into NEI as 16 recipes. 15 for white->non-white bag, one for each dye, as well as one for oredictBag + Bonemeal = white bag. That one cycles between all of the bags, including white even though that recipe doesn't work. I don't think it'd be a problem, no one should even try to do that recipe even though it technically says that you can. (it cycles between all colours of bags, spotting white would take a while).


sinkillerj:
Dust repairing will absolutely be changing.

Lilylicious:
In that case, I won't be changing the recipe file or adding them to NEI. Will revisit it after it's changed.

PR #980, opened 2015-08-29
[Done]MineTweaker galore

Lilylicious:
Preliminary pull request to invite y'all to look through it.

@MaPePeR, I added the EMC mapping late at night in a spot that is functionally working. I can move it to a better spot if you so wish, it's in your territory. Just point me in the right direction and I'll do it (tomorrow).

Other than that, I think this should be fully functional. Positive. 100%. Well, 95%. 74% positive. IT WORKS IN MY TESTS OKAY


MaPePeR:
@Lilylicious I'd rather see a separate `IRecipeMapper` for the Klein Star Recipes.

Lilylicious:
I figured something like that, will take care of it today

Lilylicious:
The changes requested by Blub are in there now, he approves.

Regarding the custom recipes I'm not 100% happy with them, but I don't think that fixing them is in the scope of this PR. Specifically I'm considering renaming RecipeShapedKleinStar since it's not really only applicable to klein stars. Maybe rename both RecipeShapelessHidden and that one to RecipePEShape(d/less)? That way it's more obvious what they do, since they're both essentially just Shapeless/Shaped custom recipes that allow us to preserve meta-data for klein stars and to hide them easily from NEI unless specifically added to it.

Lilylicious:
Threw a ton of mods into my testing environment, tried just about everything I could think of.

Beyond the annoyances in making minetweaker recipes (item names and other gripes), there's nothing I could find in our code that's producing any bugs.

I'm giving this my stamp of done-ness.

PR #984, opened 2015-08-30
[DONE] "per Conversion"-Arithmetic for fluids

### LINKED ISSUE #886: Non-vanilla liquid emc off
So i noticed in a couple different packs (madpack 2 and void) that the emc value for the liquid (melted) versions of metals is off.  The liquid version for a bucket of iron would be the cost of the bucket + the cost of one iron or 1024 emc.  The problem with this is that tinkers construct and other mods have the value of one ingot as about 1/6 of a bucket (144 mb out of 1000mb).  This can lead to a emc bug where you create a bucket of the liquid metal then convert the liquid back into the ingot form creating 6 ingots at the cost of 1 ingot.  I dont know if this is caused by the packs or not so i thought id report it anyways just in case.

MaPePeR:
So we can calculate conversions without the normal "rounding down"-stuff.

~~Still needs the actual fluids conversions... and probably a rewrite... the class hierarchy was kind of destroyed by this attempt.~~

Fixes #886


MaPePeR:
Tested with:

* `CofHCore-[1.7.10]3.0.3-303`
* `EnderIO-1.7.10-2.2.8.349`
* `Mantle-1.7.10-0.3.2`
* `TConstruct-1.7.10-1.8.5`
* `ThermalFoundation-[1.7.10]1.2.0-102`
* `MineFactoryReloaded-[1.7.10]2.8.0-104`

So this now makes all the buckets get a proper emc value instead of the "1 ingot yields 1 full bucket" we had before.

MaPePeR:
Use #1015  instead

MaPePeR:
Reopening, so it can get merged without #1015. 

Both PRs have some refactoring & cleanup in common.

Issue #989, opened 2015-09-03
Some items not calculating correct EMC when ingredients have custom values


Havanacus:
Using ProjectE-1.7.10-PE1.8.0    Forge 10.13.4.1492
I've been adding EMC values to the base ingredients in Simply Jetpacks and Redstone Arsenal.  Everything in Simply Jetpacks calculated up flawlessly once I added the raw ingredients, but Redstone Arsenal is giving odd, inconsistent values.

For example, Fluxed Armor Plating is worth 19,370 and Flux-Infused Helm is simply 5 of these crafted together.  There are no alternate recipes, yet the Helm was calculated to be worth 213 EMC.
The kicker is that the values aren't even consistent.  The boots, which require 4 plates, are calculated at 2,318 EMC.  The Legging, which require 7 plates, are calculated at 135,590 EMC.  And the Chestplate, which requires 8 plates, is calculated at 4,636 EMC.

Restarting my Client and the reloadEMC command did not help.

MaPePeR:
It might think the items are damaged and reducing the EMC value. 
Could you generate a pregen file and check what values were written into that for these items?

Havanacus:
Redstone Arsenal seems to use a separate Durability and Charge value, so the items always have a full 400,000 durability.  In any case, I spawned an uncharged helmet and charged it, the EMC value remained constant.

I generated a pregen file and the values in it appear to be correct.
"RedstoneArsenal:armor.helmetFlux|0": 96850
"RedstoneArsenal:armor.plateFlux|0": 154960
"RedstoneArsenal:armor.legsFlux|0": 135590
"RedstoneArsenal:armor.bootsFlux|0": 77480

MaPePeR:
This indicates that there is indeed something wrong with the "substract emc from damaged stuff"-calculation.

MaPePeR:
I migrated the fix from #992 into its own pr #999, so we can release it sooner. I tested it with redstone arsenal and it appeared to work fine.

Issue #985, opened 2015-08-31
Easy way to prevent exploits - Suggestion


Beelzaboss:
Hi there,

I have a suggestion regarding the EMC of ProjectE... Would it be possible to make it so that when you transmute something into EMC you will get less for it actually is, - and also add configuration options that is disabled by default... Let me try and explain bit more...

    # Should Transmutation into EMC give less  that what it is? [default: false]
    B:TransLessEMC=true

    # By how much in percentage should this occur?,  TransLessEMC needs to be true [default: 5]
    I:enableCollector=5

Example:

I as player have 3 Iron Ingots, that will be a total of 768 EMC... And when I transmute this into EMC it will not give me 768 but the value of C;

A = 768
B = 5
C = ?

C = A - (A * (B/100))
C = 768 - (768 * (5/100))
C = 729.6
C =/ 729

I personally believe this will help and prevent most possible exploits right now. And it would also be nice if another configuration option could be added that will do the following;

    # Should items with EMC value of 1 not able able to converted into EMC but still given as a result item? [default: false]
    B:EMC1Disable=true

Meaning, That I can't Transmute a Item with EMC value of 1 to EMC, but I can still get this item through EMC gathered from other items.

I hopes this makes sense!

Thanks for such a great mod!

Issue #618, opened 2015-04-04
[Feature Request] Inequivalent Exchange


Brycey92:
From @zortag:
"The original EE2 had a "%cost" option for bypass abuses on servers (cobblegen in relay). evalutate to add it (ie: cost 10% = a 100emc item cost 100 to be taken from tablet but give 90 when burned), decimal are always rounded to the lower (ie: cobblestone = 1emc burned at 10% cost gives 0,9emc, so nothing is added to the tablet emc amount)"

This proposed feature would promote trade with other players on multiplayer servers, rather than them making everything out of EMC. There could be an optional global % cost, and overrides per-item or oredict entry.

Kolatra:
Can you provide more info on this? I have looked and there was no option for this, it was most likely a server specific mod.

Brycey92:
I haven't actually seen the config option zortag referenced, but the general concept remains the same. A percentage of EMC could be added to the cost when creating, or subtracted from the EMC received when destroying. The config value could either be that percent to add to the base value of 1.0, such as -0.1 or 0.1, or it could be a multiplier, such as 1.1 or 0.9.

Kolatra:
I will talk to sinkillerj and we may consider something like this.

zortag:
sorry for my late answer.
i actually have no 1.2.5 pack on my computer, so i can't check the config file.
i started making modpacks for personal use with that version but, except the big work to solve id conflicts, i don't remember much about 1.2.5 modpacking.
i remember i found that option, maybe it was a feature of any ee2 side mod. i used much customemc to add emc values to item not recognized by the main mod

Shalashalska:
I implemented something like this in a fork I made a few months ago. It's quite simple really, just add a config option that sets static variables, which incoming EMC is multiplied by. About 20ish lines of code, and no effects if the default is set to 1, but very nice for people who want a major nerf to transmutation without removing it completely.

MaPePeR:
I think, we would appreciate a pull request.

Brycey92:
@williewillus Was this feature implemented or canceled? I see the PRs are closed, not merged, but that was over 2 years ago so I'm not sure if you implemented it yourselves in the mean time. Please let me know!

williewillus:
this was implemented by sin in one of the 1.12 releases

Brycey92:
Any possibility of a backport to 1.10?

Issue #784, opened 2015-06-09
[Suggestion] Ability to add EMC to items but not able to learn the item


Carrots4U:
I'd love to be able to give an item an EMC value (for example, manasteel) and be able to put it into a table and get the EMC, but not be able to learn/transmute back into the item. This would be very helpful in balancing packs around Project E.

Carbon-Ghost:
+1

sinkillerj:
I plan to split buying and selling into two methods so that the sell price can be less than the buy price when transmuting, when that is introduced I will consider this as it should be fairly simple to implement once the methods are split.

Tencao:
I too would love this idea to be added. It's not major, but something would also help servers massively.

pupnewfster:
This is possible via external addons using the events merged from pull #1707. I know for a fact that they can be locked this way via Compat Skills (what I had originally needed it for), not sure if any other addons use these events yet.
If having it so addons can do this is enough to meet this issue it can be closed, otherwise it should be kept open.

Tencao:
I would prefer to be a tag on the item rather than via addon, as we're then at the mercy of addons adding compat for all items used by this mods emc system

pupnewfster:
The different buy/sell cost thing is not possible, realizing now that may not have been clear. Though I do see how having a built in thing that uses those events to stop items from being learned, while allowing them to keep an EMC value would be useful.

PR #992, opened 2015-09-04
Split buy and sell EMC Values & Multiplier

### LINKED ISSUE #594: Tooltips: show durability EMC loss/stored EMC gain
Maybe the tooltips should show a bit more how the actual value is calculated.

Like: 

`BaseValue + StoredEMC + EnchantBonus- DurabilityLoss` ...

### LINKED ISSUE #988: Duping tools with enchantments
As a player has reported to me, duping tools with enchants doesn't work. It asks for the tools emc + enchant emc, but it only outputs normal tools. On a side note, and maybe related to this, the reported tools duped by this would not repair. The repair part could be my doing, but I'm leaving it here incase it's related.

MaPePeR:
# Don't merge this!
## Merge #1015 instead, which also contains #984

With this PR:
* Fixes #988 
* Split a lot of calls to `getEmcValue` and `doesHaveEmc` into `getEmcValueForDestruction`, `getEmcValueForCreation`, `hasEmcValueForDestruction` and `hasEmcValueForCreation`. 
* Implements: #985, #618 Option to add a multiplier that is applied when something is converted to EMC/destroyed.   
  With Side Effects:
  * Have to use the left side or the lock slot to learn an item when its EMC has been reduced to 0. Cannot use Shift-Click to learn
  * Mercurial Eye cannot transmute Stone anymore  (Or should i not apply the multiplier for that?)
* Detailed EMC tooltips when sneaking. Shows amount removed because of durability and amount added for Enchantments. Fixes #594 (Does NOT show the Multiplier EMC yet!)

Missing:
* Fuel upgrading does not use the new methods yet, because i don't know which to use.
* Fuels in general (Should these also be reduced by the multiplier?)
* Tooltips that show the reduced EMC Value - but when and how? **Suggestions?**
* Diving Rod
* API (More API methods? Which one should replace our current?)
* Left Side of the Transmutation Table (Which items should be allowed in there?)


williewillus:
* Fuel upgrading - should use whichever value is used when taking out of the right side of the transmutation (the emc -> item value, so buy I guess)
* Fuels (I assume you mean when being consumed by power items) should yield the item -> emc value aka sell value on consumption
* Tooltips - not sure
* Divining Rod - Should show the item -> emc value (sell) since the items being divined will eventually be sold via condensing, GoED-ing (did you update GOED?), or transmutation sacrifice
* API - Add extra methods and either deprecate the old ones or write super clear doc clarifying which one it now delegates to
* Left side of transmutation - anything which has a value period (need sell to teach table, need buy to filter outputs)

MaPePeR:
Also fixes #989: It now only uses the "damage calculation" for an item, when it cannot find an emc value for the `id:metadata` pair. So it tries `id:metadata` and when it finds a value uses it. If there is no value it tries `id:0` and applies the damage multiplier using `ItemStack.getItemDamage()/ItemStack.getMaxDamage()`. (These call methods in `Item` which are overwritten by Redstone Arsenal causing the problem. Before it always applied the damage modifier regardless of metadata.)

So the damage multiplier still works for Vanilla Tools - not sure about others.

I changed the GOED to use the "item -> emc" value for condensing and "emc -> item" value when creating an item. 

With this PR it should be easy to integrate the requested #784 "unlearnable Items". Not sure if the GOED should automatically consume these, though. And how to visualize this property in the ToolTip.

I just noticed that "buy and sell" are very bad words to describe this, because its unclear if the player or the table is the one who is buying/selling.

williewillus:
Yeah, maybe it's better to refer to it as "Creation" and "Sacrifice" (maybe something more arcane sounding when I can think of a word) to refer to emc -> item and item -> emc, respectively

MaPePeR:
Besides the tooltip needing proper localisation & the API i think this is actually done now... .... ...

MaPePeR:
So what do you think: How should we do the tooltip?

> Create EMC: 123
> Destroy EMC: 456

When they differ and 

> EMC: 123

When both values are the same?

The problem is, that the "destroy" emc value is different for a lot of stuff: When you used a tool it looses emc, so the create/destroy values will not be the same anymore.

Another option would be to do (Maybe green/red colored?):

> EMC: 123/456

and if the values are the same

> EMC: 123

williewillus:
I'd say keep them separate

MaPePeR:
Then we need to agree on words.

Carbon-Ghost:
pls be color blind friendly :o

williewillus:
Creation and Destruction, or something similar or more lore-y.
I think a word like that make it pretty clear what process is happening (emc -> item and the reverse, respectively
Brain isn't working but maybe something like "Condensation" for creation since that's already in our lore, and make up anoter good word for item->emc. Dissolution maybe?

MaPePeR:
That are very long words. I would prefer to keep these tooltips short. Maybe there is a symbol or single character we can use? 
Or just do a 

> C. EMC: 123
D. EMC: 456

And only write create/destroy/the long form  when sneaking?

williewillus:
eh I feel like "to" and "from" are as ambiguous as buy and sell, I'd be down for C: and D:, with the full words on shift. Need the abbreviation to be localized separately from the main one though, just in case of weird languages

MaPePeR:
So this is how it looks right now (Default config without the multiplier)

image

**Please** give your feedback.

MaPePeR:
Closing this for #1015

PR #1000, opened 2015-09-13
[DONE] Chisel version check and no dependency

MaPePeR:
Because it only caused problems in my dev workspace i just wrapped the ChiselAPI with our own classes & used reflection to access everything.

So we now check the Version to start with "2." instead of comparing the Modname (which apparently changed from "Chisel 2" to "Chisel" recently)


sinkillerj:
This does not address the core issue of why we used the name instead, the forks still cause issues.

MaPePeR:
Do they? I talked to the chisel guys a little and they said the version should work.

If i understand it right there are 3 Chisels:

* Chisel with version 1.x 
* Chisel 2 with version 2.x
* Chisel with version 2.x

Chisel with version 1.x is the one we don't support.

And i'd rather 'enable' the mapper for chisel versions it does not support and have it throw a (catched) exception than disable it for chisel version it does support.

williewillus:
The check for "Chisel 2" as a name was because Chisel 1 and Chisel 2 (before the recent fork) had the same modid and I was a derp and didn't think of checking version numbers. Just in case you guys were wondering.

PR #1015, opened 2015-09-21
Split buy and sell and fluid conversion-PRs merged!

### LINKED ISSUE #594: Tooltips: show durability EMC loss/stored EMC gain

### LINKED ISSUE #618: [Feature Request] Inequivalent Exchange

### LINKED ISSUE #784: [Suggestion] Ability to add EMC to items but not able to learn the item

### LINKED ISSUE #886: Non-vanilla liquid emc off

### LINKED ISSUE #988: Duping tools with enchantments

MaPePeR:
Merges #984 and #992 

* Fixes #784  - **Add EMC, but not able to learn it** (Needs manual tweaking the pregen-file - but its possible!)
* Fixes #618 - **Inequivalent Exchange - %cost multiplier**
* Fixes #594 - Show durability-emc-loss & enchant-bonus in tooltip
* Should fix #988 - Duping tools with Enchants
* Fixes #886 - Non Vanilla Liquid EMC off

**Please** still comment on the Tooltips [here](https://github.com/sinkillerj/ProjectE/pull/992#issuecomment-142027898)


MaPePeR:
With the latest commits i think i have all my TODO for both pulls complete.

MaPePeR:
Now that @Lilylicious actually tested the [Conversion API](https://github.com/MaPePeR/ProjectE/tree/apiForModRecipes) i based upon this PR: Should i also add it to this PR?

MaPePeR:
@williewillus @sinkillerj I still would like to have some feedback on the tooltips and we need to decide if the tablet and condensers should accept 0-emc items or not. 

Tooltips

williewillus:
Looks fine to me
The C. D. was a little weird at first

How does it handle when someone stacks a bunch of normally unstackable items? (Stack of 32 enchanted picks together)

williewillus:
And I vote no for 0-emc items, 0 should be the special case for no value until we figure out an actual thing to use for no value.

MaPePeR:
The Tooltip for stacked Items actually was not changed at all. It always shows the destruction-emc multiplied with the stacksize. (So for the Damaged Pick that would be stacksize * 8836)

"No EMC" has always been defined by "is present in the EMC map" and the reason we have (and always had) an extra `hasEMC`-Method, i think.

The problem is, that with a multiplier < 1 (and everything else does not make sense) for destruction EMC the EMC value for Cobblestone will become 0. You will still be able to learn it and create it for 1 EMC. But should we allow you to feed it into a table, so people can get rid of it?

Maybe make that a config option to disable/enable that behavior?

sinkillerj:
I made the guidelines for 0 emc handling and sell methods clear.


There is no reason to pregenerate sell values, nor to put them in the tooltip. In the config file should be a simple sell ratio, when a item is burned the sell method is called, emc x ratio rounded down, if the rounded value == 0 it becomes 1 instead. 


Individual sell values will not be accepted into master, this adds unnecessary complexity and confusion. Ideally the covalence loss would be displayed as a percent in a small separate window on devices which can burn items, but only if the ratio is not 1. Example, if the sell ratio is 0.8 the window would read: "Colalence loss: 20%"

MaPePeR:
Unlearnable/Unconsumeable Items have been requested several times and getting from the %cost/multiplier to that is only a very tiny step. The methods you need in the API/EMCHelper even stay the same.

Sell values are not pre"generated" in a very special way. They are calculated with the multiplier. When pregenerated emc is enabled they are written to the pregen-file - just like the "normal" values.

>  if the rounded value == 0 it becomes 1 instead.

The multiplier was requested especially to have the destruction-emc value of cobblestone = 0, to reduce the efficency of cobble gens.

sinkillerj:
If they want a 0 emc sell value they will need to make a fork, there are no plans to allow this.

MaPePeR:
I should have been more specific: 

> The multiplier was requested especially to have the destruction-emc value of cobblestone = 0 **OR no destruction-emc for cobblestone at all**, to reduce the efficency of cobble gens.

The alternative to a 0 emc cell value is "no emc sell value"...

Lilylicious:
Throwing in my unsolicited opinion, I don't see the harm in allowing an item to be created but not destroyed. A way to inhibit cobblegens would play a big part in balancing the mod.

sinkillerj:
The debate here is not buyable and sellable flags, those are still planed, its individual sell values.

MaPePeR:
Having a destruction emc value, that differs from the creation emc value (even if caused by a simple multiplier), should be presented as 2 values to the user. (Noone wants to do the math oneself, to know how much one gets from burning an item) 

Do we agree on that?

MaPePeR:
@sinkillerj so i would suggest that instead of discussing this now we should merge #1036, which contains the infrastructure and fixes a few bugs related to this. 
This would give Mordenkainen/Other API Users  time to adapt to the new methods that are needed and us more time to discuss how we want to implement this if at all.

PR #1035, opened 2015-10-03
Collectors accept emc from providers when upgrading/charging. 

### LINKED ISSUE #1021: Relays don't push EMC to active Collectors in power flowers
As title says. I have 5 Relays around a central Collector, and Collectors around the Relays (6t 6w 5d design), and what it USED to do is pull energy from the Relays, which was being pushed by the other Collectors, to speed up the passive upgrade of EMC-fuels in the center Collector.

I mean no offense by this, but it looks like someone new took over the coding side of things.. so, for reference, it was working fine in version 1.5.0. I took a break from Minecraft and that was the last one I'd had installed.

EDIT: Downgraded to version 1.8.0 from 1.9.2 and the issue is resolved. Dunno how many internal versions there were between these, but I hope it helps narrow down the search some.

MaPePeR:
Fixes #1021

(Has the side effect, that when you take the fuel out the EMC stored in the collector for upgrading will be drained back into the relays, but i think that is not a problem)

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/tiles/CollectorMK1Tile.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/tiles/CollectorMK1Tile.java
@@ -2,6 +2,7 @@ package moze_intel.projecte.gameObjs.tiles;
 
 import cpw.mods.fml.common.network.NetworkRegistry.TargetPoint;
 import moze_intel.projecte.api.item.IItemEmc;
+import moze_intel.projecte.api.tile.IEmcAcceptor;
 import moze_intel.projecte.api.tile.IEmcProvider;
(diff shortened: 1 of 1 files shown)


williewillus:
Not sure if this is an actual problem, but test it with relays and make sure it doesnt have infinite loops or infinite bonus exploits

MaPePeR:
This is mostly to get the power flower with collector/fuel upgrading core working. 

The send to relay bonus is not applied when accepting emc. 
I tested it with the power flower with relays and could not see a problem. 
It only accepts emc when its upgrading  fuel and only provides emc when it does not, so there should be no infinite loop. (having fuel to upgrade toggles between being an acceptor/provider)

MaPePeR:
I can't think of a way to exploit this.

PR #1036, opened 2015-10-04
Split create and destruction EMC Methods. API/Methods only

MaPePeR:
Also contains #984 

Cherry picked some commits from #992/#1015.

* Splits calls to `getEMC` into a version for creation and a version for destruction.
* Adds API methods to check for creation and destruction EMC values.

The only difference is, that destruction EMC honors enchantment bonus and destruction multiplier for now.


MaPePeR:
I think we even might need another method that does `ItemStack -> ItemStack` like`getCreatedItem`, which does the nbt_whitelist lookup and returns an undamaged tool when given a damaged tool. Thoughts? (Still think we can merge this how it is right now)

PR #1039, opened 2015-10-04
Allow unlearning from full knowledge

### LINKED ISSUE #970: Can not unlearn anything which is learned by tome of knowledge

Lilylicious:
I decided to check to see how the performance was impacted by getting rid of the full knowledge cache and replacing it with addKnowledge for each item not currently known. As it turns out, the difference is noticeable but not inhibitingly so. There's a tiny moment of lag right as you learn the tome. After that, what you're left with is a map from which you can remove items, allowing players to unlearn items even with a full set of knowledge.

The only reason I can think of that could be an argument against this is the performance impact, which like I mentioned is minimal.

Would close #970

### DIFF
--- a/src/main/java/moze_intel/projecte/api/proxy/ITransmutationProxy.java
+++ b/src/main/java/moze_intel/projecte/api/proxy/ITransmutationProxy.java
@@ -47,7 +47,9 @@ public interface ITransmutationProxy
      * If called on the client side, playerUUID is ignored and the client player is used instead
      * @param playerUUID The Player to query
      * @return Whether the player has full/override knowledge from the Tome, false if player is not found
+     * @deprecated as of API version 8, use {@link #hasKnowledgeFor(UUID, ItemStack)}
      */
+    @Deprecated
     boolean hasFullKnowledge(UUID playerUUID);
 
     /**
--- a/src/main/java/moze_intel/projecte/gameObjs/container/inventory/TransmutationInventory.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/container/inventory/TransmutationInventory.java
@@ -63,7 +63,7 @@ public class TransmutationInventory implements IInventory
 		{
 			learnFlag = 300;
 			
-			if (stack.getItem() == ObjHandler.tome && !Transmutation.hasFullKnowledge(player))
+			if (stack.getItem() == ObjHandler.tome)
 			{
 				Transmutation.setFullKnowledge(player);
 			}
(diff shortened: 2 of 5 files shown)


williewillus:
I wouldn't remove the entire pre-calculated cache map, keep it there and just use addAll on it or something
So it wouldn't have to be recalculated for every single player
This'll also need to be notified to API users - its a pretty major change in terms of how "full knowledge" is handled now.

Lilylicious:
Good point. To avoid a cluttered map, in the addFullKnowledge part I'll just clear knowledge and set player knowledge = the pre-calculated cache

MaPePeR:
Make sure to copy said full knowledge, so the players don't modify the full knowledge for all other players.

Lilylicious:
I redid the entire thing, keeping the cache this time. I don't think there should be any problems, though I'd appreciate y'all taking a look through it and checking. I add full knowledge by clearing the getKnowledge list and using addAll(cache) on it.

MaPePeR:
>  I add full knowledge by clearing the getKnowledge list and using addAll(cache) on it.

Yea, thats the right way to do it.

>  set player knowledge = the pre-calculated cache

would be a bad idea ;)

Lilylicious:
I'm not entirely familiar with how lists interact in Java, my intuition tells me that List = AnotherList would copy AnotherList to List, leaving two separate lists with identical contents.

My experience however.. makes me wary.

MaPePeR:
List = AnotherList would copy the reference, so the lists become the same list.

So after `a=b` `a.add(something)` would also add `something` to `b`. Because both variables reference the same list.

I think we can get rid of `hasFullKnowledge`. Its not really needed anymore and with the change it has a horrible complexity.

Lilylicious:
The reason I kept it in this iteration is that it allows us to keep that check, which is useful in some edge cases on our end and doesn't mess up any API usages that may be present.

This iteration has no noticeable impact on performance, so I don't see the harm in keeping it. There may be ways of optimizing it however, but I'm not sure.

I considered making the API call deprecated and return false, but that's not entirely accurate. Could also stop using `hasFullKnowledge` internally and keep it in the API call.

MaPePeR:
I think there is no real usage for `hasFullKnowledge`. Not even in the API. I think deprecated & false default is fine.

If we *really* want to improve knowledge we should make it `Map<ItemId, List<ItemStack>>`, so we don't have to iterate over ALL the elements to check if an ItemStack is present in the knowledge. (And that probably would have a measurable performance impact)

Lilylicious:
The map or the iteration over elements? I could probably tinker around with knowledge some more and see if I can't improve it, either with a map or some other method. Maybe get some input from the two silly time zoners that aren't here? @williewillus @sinkillerj

MaPePeR:
I meant: Switching from iterating over all the items to only iterating the items with the same itemId would probably have a measurable performance impact.

Lilylicious:
Perhaps, though I'm not having any noticeable negative effect from the checks. The processes that use the iterator appear instantaneous.

sinkillerj:
I agree with the removal of hasFullKnowledge, its pretty silly.

Lilylicious:
`hasFullKnowledge`  has been removed.

PR #1041, opened 2015-10-05
Fix smaller/smaller equals derp

MaPePeR:
So before you couldn't pull out an item for which you had the exact amount of emc.


sinkillerj:
And somehow we both missed this when I merged it -,-

MaPePeR:
When i woke up the other day one of my first thoughts was "Hey, that code you wrote... did you check if you used the right comparison operator..." turns out i didn't... :smile:

PR #1056, opened 2015-10-16
Make Fluid World Transmutations one-way only

MaPePeR:
Because they do not work, so we don't want them to show in NEI.

Untested.


williewillus:
why 0.o
they work in 1.8

williewillus:
they work in 1.10 now

PR #1066, opened 2015-10-25
Check for overflow on EMC display value

### LINKED ISSUE #1065: MK2 Condenser breaks when consuming slot contents with EMC > MAX_INT

kenzierocks:
Shows the required EMC instead of the current EMC when the current EMC is greater than `Integer.MAX_VALUE`.

Fixes #1065.


pupnewfster:
I think this can be closed as it is out of date, and #1780 is a better (and updated) solution.

PR #1068, opened 2015-10-27
Consistency changes

Samario1:
The normal message is "N has joined the game", not "N has joined the server"
And infinite water bucket / infinitely full lava bucket, changed the prior


sinkillerj:
At this time the decision has been reached to not accept these changes, the current wording has been deemed more appropriate.

PR #1074, opened 2015-10-30
Keybind to toggle Aura Node display.

Kolatra:
As requested by MattaBase.


williewillus:
Depending on the way the thaumcraft interface methods are called, not sure if this will actually work. Even if it did, static fields in item classes are generally a super bad idea because items are stateless (only in extremely rare circumstances would you want it to apply globally)

Kolatra:
I did test this actually, it does work.

I also used the code already existing as quite the reference for this implementation.

sinkillerj:
As this feature is being moved to another mod I am not implementing any further changes prior to its removal, something I informed you of before you made this pull.

williewillus:
@Kolatra where else in the codebase are static fields in item classes being used? if there are any I need to zap them

PR #1080, opened 2015-11-01
[DONE] Fix NEI Hotkeys in Transmutation GUI

### LINKED ISSUE #1057: can't show NEI-recipes/uses while having the transtable/transtablet open
when pressing 'R' on a item, it doesnt show the item, this only occurs while in the gui of the transmutationtable or the transmutationtablet

MaPePeR:
Fixes #1057

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/gui/GUITransmutation.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/gui/GUITransmutation.java
@@ -117,7 +117,7 @@ public class GUITransmutation extends GuiContainer
 				inv.updateOutputs(true);
 			}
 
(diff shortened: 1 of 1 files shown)


MaPePeR:
Anything wrong with this PR? @sinkillerj

sinkillerj:
This could cause issues with the filter box and IMO the issue is of non importance.

MaPePeR:
I tested it and it worked. Did you test it and encounter any problems?

MaPePeR:
Surely this is not a major game breaking bug, but it is an annoyance that we can fix it, so why shouldn't we? Especially, if we want to call it finished sometime. 

If you look at the changes of this PR, you can see, that if `this.textboxFilter.isFocussed() == true` the instructions that are run are 100% identical. 

If it is not true it just defaults to `super.keyTyped`, instead of checking for the escape and inventory key ourself.

williewillus:
I guess this doens't really matter anymore since NEI is ded

PR #1107, opened 2015-12-07
[DONE] Fix Double-Click Exploit #1104

### LINKED ISSUE #1104: Major EMC exploit - Transmute tablet
Ok, so this was reported by a player, and had difficulty explaining it, until he showed me a video

http://i.imgur.com/KkphYKf.png

https://www.youtube.com/watch?v=HJS6OpBsfc4

MaPePeR:
Check if requested emc is actually available in Slot.decrStackSize

Also calculate the proper EMC value without ignoring the requested stacksize

Fixes #1104

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,7 +43,7 @@ repositories {
     }
     maven {
         name 'Chisel'
-        url "http://coloredlightscore.us.to/maven/chisel-2-dev/"
+        url "http://maven.tterrag.com/"
     }
     ivy {
         name "MineTweaker3"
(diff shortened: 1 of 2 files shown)


djoveryde:
Does anyone have a copy of this with the patch that I can add to my server?

MaPePeR:
Apparently this PR does not fix the issue :cry: 

I'm out of ideas.

MaPePeR:
Had another idea, that *i think* removes the exploit, but still leaves some client/server desync. Its still A LOT better than without these commits.

PR #1128, opened 2016-01-01
Added the ability to set EMC values on ItemStacks with NBT data

solidDoWant:


src/main/java/moze_intel/projecte/emc/SimpleStack.java
```diff
@@ -76,6 +82,9 @@ public boolean equals(Object obj)
 			}
 
 			//return this.id == other.id && this.damage == other.damage;
+			if(this.nbt != null && other.nbt != null){
+				return this.id == other.id && this.qnty == other.qnty && this.damage == other.damage && this.nbt.equals(other.nbt);
+			}
 			return this.id == other.id && this.qnty == other.qnty && this.damage == other.damage;
```
MaPePeR: This return should only happen, when `this.nbt == null && other.nbt == null`, i think. (The Stacks are not the same, when one has NBT and the other has not!)

MaPePeR:
I hope you tested this a lot

williewillus:
^ yes please.
Write a bunch of unit tests for this in the tests/ folder

MaPePeR:
Not sure if you can actually unit test this, though...

src/main/java/moze_intel/projecte/emc/SimpleStack.java
solidDoWant: Should be fixed in this last commit.

solidDoWant:
Not sure how to test it with the project without adding a bunch of dependencies/messing the whole thing up. I could fork this branch though and do it if you'd like

src/main/java/moze_intel/projecte/api/proxy/IConversionProxy.java
```diff
@@ -21,7 +21,7 @@
 	 *
 	 * You can use the following things for the {@code output}-Parameter and the keys in the {@code ingredients} Map:
 	 * <ul>
-	 *     <li>{@link ItemStack} - The ItemId and Metadata will be used to identify this ItemStack (May contain a {@code Block} or {@code Item}). You can use {@link net.minecraftforge.oredict.OreDictionary#WILDCARD_VALUE} as metadata.</li>
+	 *     <li>{@link ItemStack} - The ItemId and Metadata, and NBT data will be used to identify this ItemStack (May contain a {@code Block} or {@code Item}). You can use {@link net.minecraftforge.oredict.OreDictionary#WILDCARD_VALUE} as metadata. If you don't want NBT data to be compared, set it to null.</li>
```
kenzierocks: `The ItemId and Metadata, and NBT data` -> `The ItemId, Metadata, and NBT data`

src/main/java/moze_intel/projecte/emc/EMCMapper.java
```diff
@@ -141,13 +141,13 @@ public static void map()
...
 					int id = Item.itemRegistry.getIDForObject(obj);
-					emc.put(new SimpleStack(id, 1, normStackItem.damage), entry.getValue());
+					emc.put(new SimpleStack(id, 1, normStackItem.damage, normStackItem.nbt), entry.getValue());
 				} else {
 					PELogger.logWarn("Could not add EMC value for %s|%s. Can not get ItemID!", normStackItem.itemName, normStackItem.damage);
 				}
 			}
 		}
-
+
```
kenzierocks: plz

src/main/java/moze_intel/projecte/emc/NormalizedSimpleStack.java
```diff
@@ -143,7 +156,13 @@ public boolean equals(Object obj) {
 		@Override
 		public String json()
 		{
-			return String.format("%s|%s", itemName,  damage == OreDictionary.WILDCARD_VALUE ? "*" : damage);
+			if(this.nbt == null){
+				return String.format("%s|%s", itemName,  damage == OreDictionary.WILDCARD_VALUE ? "*" : damage);
+			} else{
+				System.out.println("JSON " + String.format("%s|%s@%s", itemName,  damage == OreDictionary.WILDCARD_VALUE ? "*" : damage, nbt.toString()));
```
kenzierocks: remove debug printing?

src/main/java/moze_intel/projecte/emc/NormalizedSimpleStack.java
```diff
@@ -254,15 +273,23 @@ public String toString() {
...
 		String itemName = serializedItem.substring(0, pipeIndex);
-		String itemDamageString = serializedItem.substring(pipeIndex + 1);
+		String itemDamageString;
+		String itemNBTString = null;
+		if(atIndex < 0){
+			itemDamageString = serializedItem.substring(pipeIndex + 1);
+		} else{
+			itemDamageString = serializedItem.substring(pipeIndex + 1, atIndex);
+			itemNBTString = serializedItem.substring(atIndex + 1);
+		}
```
kenzierocks: Could use `JsonToNBT` in the minecraft source here. That will do the deserialize for you.

src/main/java/moze_intel/projecte/emc/SimpleStack.java
```diff
@@ -76,7 +82,13 @@ public boolean equals(Object obj)
 			}
 
 			//return this.id == other.id && this.damage == other.damage;
-			return this.id == other.id && this.qnty == other.qnty && this.damage == other.damage;
+			if(this.nbt != null && other.nbt != null){
```
kenzierocks: May want to ignore NBT for comparison if either are `null` right now, to maintain old behavior.

solidDoWant: If either are null they're not the same item, so it returns false.

src/main/java/moze_intel/projecte/emc/NormalizedSimpleStack.java
```diff
@@ -116,17 +123,23 @@ public static NormalizedSimpleStack forOreDictionary(String oreDictionaryName)
...
+		
+		private NSSItem(String itemName, int damage){
+			this(itemName, damage, null);
 		}
 
 		@Override
 		public int hashCode() {
-			return itemName.hashCode() ^ damage;
+			return itemName.hashCode() ^ damage ^ (nbt == null ? 0 : nbt.hashCode());
```
MaPePeR: Does nbt implement the hash code function properly? If it falls back to the object implementation this will cause problems, because objects hash method uses references.

solidDoWant: Taking a look at how minecraft deals with nbt hashes,

NBTTagCompound's hashCode calls super(NBTBase)'s hashcode, and Map's hashcode.

NBTBase's hashcode calls this.getID(), which is abstract, so when NBTTagCompound.hashCode() is called, NBTTagCompound.getID is called. getID returns (byte)10.

According to Map's javadoc, "The hash code of a map is defined to be the sum of the hash codes of each entry in the map's entrySet() view". Taking a look at the documentation in NBTTagCompound, the Map instance's key is of type String. The String class's javadoc says "The hash code for a String object is computed as
 s[0]*31^(n-1) + s[1]*31^(n-2) + ... + s[n-1]".

None of the classes called when nbt.hashCode() is called call Object.hashCode().

MaPePeR: Good.

MaPePeR:
It would also probably be wise to copy the nbt when saving it in a (normalized) simple stack, if you are not already doing that. 

It's very important, that the normalized simple stacks stay immutable.

solidDoWant:
Taken care of. If I didn't it'd cause all kinds of problems.

I didn't change the mutability (is that a word) of anything.

solidDoWant:
Finished writing and testing my Mekanism customConfig file. Everything works 100%.

https://github.com/solidDoWant/ProjectE-Custom-Conversions/blob/master/Mekanism.json

Malkierian:
Really sad this hasn't been integrated yet.  Was just playing a pack with Mekanism in it, and the energy cubes rely on NBT to differentiate (which is weird, because almost all of the rest of the mod uses metadata), so I can't have all four cubes in there.  I think the factories also use NBT.

pupnewfster:
I believe this can be closed in favor of #1858 being newer and seemingly more detailed.

pupnewfster:
Closing this as it is rather out dated and I have a *mostly* finished [implementation](https://github.com/pupnewfster/ProjectE/tree/nbt_nss) of NBT. But also #1877 is a more modern version of this (which I am going to leave open until I finish with my branch)

PR #1160, opened 2016-01-25
Prevent Consume from Unlearn Slot (Dupe Fix)

Robotia:
A player on my server found a dupe glitch where consuming items from the unlearn slot allows them to produce free EMC.

I wasn't able to replicate this myself, the entities kept disappearing for me, but here is the video of him doing it on my server :https://www.youtube.com/watch?v=OYIhOrFaCLU&feature=youtu.be

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/container/TransmutationContainer.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/container/TransmutationContainer.java
@@ -105,7 +105,7 @@ public class TransmutationContainer extends Container
 			
 			transmutationInventory.updateOutputs();
 		}
-		else if (slotIndex >= 26)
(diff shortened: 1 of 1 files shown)


Robotia:
@sinkillerj ?

MaPePeR:
Can you explain the fix and what causes the dupe?

Robotia:
Basically, unlearned items can be shift-clicked from the unlearn slot into the consume slot. This prevents that by raising the inventory index to 27. Index 27 and above is the player's inventory.

MaPePeR:
I kind of disagree with this kind of fix. Because allowing to shift click from the unlearn slot is not the reason the dupe exists.

MaPePeR:
This was instead fixed with #1220

Issue #823, opened 2015-06-30
In game guide book


Lilylicious:
Opening the issue to keep track of guidelines/requests/thoughts.

1. No mod dependency.
2. Index support.
3. Basic image support.
4. Attempt to keep page adding/sorting/management as easy and streamlined as possible.

williewillus:
Perhaps a way to read from an online resource such as a git repo (so we can make live edits)

Maybe some way to incorporate the current wiki (which i should get back to work on)

sinkillerj:
I dont see the need for online reading simply because any text is going to be specific to the version of the mod and what it includes.

Lilylicious:
Online reading adds another layer of complexity that is frankly unnecessary. The main question for me is the approach: Menu or text heavy.

Botania is an example of a menu-heavy book, where you click a category that contains pages that you click to open. It makes the guide less limited into a traditional book structure and unlocks more freedom in what type of pages you add, whether it be text or big pictures or recipe pages and whatnot.

RFTools on the other hand follows a much more text-heavy structure and is more akin to a real book, where page 2 follows page 1 etc. With a table of contents with links and a searchable index, this can be as easy to navigate as the menu based system while allowing for a more natural book feel.

sinkillerj:
I would like to see the index be like:
**Category Name**
Link
Link
Link

sinkillerj:
This may result in the index being multiple pages, but I dont feel that each category needs to link to its own page of links.

Lilylicious:
Do we want to keep pages linked together traditional book style? Turning pages in other words. Could be useful for misclicks, simply turn the page to the right one.

sinkillerj:
Yea the index just linking to a page number would be nice so you can flip through, but you could always click "return to index" at any time.

Lilylicious:
@williewillus, regarding the location for the files: If we want to keep it gathered, it can all go in gameObjs.items.manual or something to that nature. Another option is to divide up the code for the manual (previously mentioned package) and put pages somewhere else, perhaps even in assets depending on how I implement things.

williewillus:
Just do whatever for now, I find it easier to reorganize as it develops than trying to factor it all in at the beginning before anythings written :)

Lilylicious:
After doing research and thinking, I've decided on how to proceed (for now) in regards to the page infrastructure.

There will be a directory of txt files from which the mod will grab and fill arrays. These will then assign their index to a series of constant ints to make the code more readable and fluid.

[...]

PR #1172, opened 2016-02-06
Fix#1171

### LINKED ISSUE #1171: Crash when switching language
Trying to switching Minecraft language from English (US) to simplified Chinese （简体中文－中国），then the game will freeze for a while and crash. I have tried removed ProjectE and after that I succeeded switching language.
[The crash-report is here](http://pastebin.com/MDAikdSV), using Forge 10.13.4.1614 and ProjectE 1.7.10-PE1.9.5.
I have no idea why there is a stackoverflow when I switch language, but according to the crash-report I think I need to report here first. 
If anyone need more information please let me know. Also if I make any mistakes please point it out.

ShieLian:
Add judgement of c0=='\s' to avoid the infinity invoke caused by space.
Close #1171 .

### DIFF
--- a/src/main/java/moze_intel/projecte/manual/ManualFontRenderer.java
+++ b/src/main/java/moze_intel/projecte/manual/ManualFontRenderer.java
@@ -37,7 +37,7 @@ public class ManualFontRenderer extends FontRenderer
         {
             String s1 = str.substring(0, j);
             char c0 = str.charAt(j);
-            boolean flag = c0 == 10; // Changed here: Remove check for space (ascii 32)
+            boolean flag = c0 == 10||c0==32;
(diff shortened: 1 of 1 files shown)


williewillus:
That check was removed for a specific reason (which I don't remember)

Check that the other languages still look fine in the book.

ErrorCraftLP:
https://github.com/sinkillerj/ProjectE/issues/823#issuecomment-127085351

ShieLian:
Opps...It's a tricky problem. The FontRender seems to be designed for English or other language the split  words with space, however, Chinese doesn't need any space. So it causes that there are no spaces in a sentence, which leads to the result: any space may cause the endlessly invocation....... So, the easiest way to solve the problem is delete all the space in zh_CN.lang. I'll do that in my brunch.

PR #1174, opened 2016-02-07
Mc18

### LINKED ISSUE #1171: Crash when switching language

ShieLian:
Delete some useless space in zh_CN.lang to avoid crash.
Close #1171.


ErrorCraftLP:
You need to revert the change to the FontRenderer ;)

ShieLian:
@ErrorCraftLP It's reverted :P

williewillus:
just curious - does this mean that all Asiatic lang files can't have spaces in them?

ShieLian:
@williewillus Actually, I know nothing about other Asiatic language. But for Chinese(both simplified and traditional), just the part of text in the guide book need remove the space in the sentence which is long enough to take up a line.(this case will cause the stack overflow) But for making sure, all space in the text of guide should be deleted or replaced by ',' or something else.

3TUSK:
Sorry if this message bothers anyone. 
@williewillus AFAIK, Chinese and Japanese has no use of space *grammatically*, cause both of them still use [Chinese characters][wikipedia:Chinese Characters], which should not cause anything ambiguous. Korean doesn't use Chinese characters under most circumstances, so they need to use space to distinguish between homonyms. 

Most languages on the word uses sort of [alphabets][wikipedia:alphabet] (e.g. latin, cyrillic, arabic) as their scripts, and thus most of them use spaces to divide between words. Examples in Asiatic languages: Vietnamese, Taglog.
Anyone who find I make mistake(s), please point it(them) out.

By the way, sometimes i18n is not easy. Take botania as example: you will find that there *are* lots of space in its Chinese translation, especially in lexica botania part, but all of those spaces are used for starting a new line - otherwise, sentences in lexicon will be rendered as only one line and be out of GUI, or even out of the screen. 

[wikipedia:Chinese Characters]: https://en.wikipedia.org/wiki/Chinese_characters
[wikipedia:alphabet]: https://en.wikipedia.org/wiki/Alphabet

PR #1207, opened 2016-03-26
Reduce unnecessary calls to getItemDamage()

Robotia:
On my server, I have a mod that makes excessive calls to insertion and removal from a Condenser's inventory. Because of this, I noticed a node (though small) for `getItemStackDamage()` under `EMCHelper`.

To get rid of this slowdown, I have restructured the constructor for SimpleStack so that it can be initialized without a damage value. This is done in the EMCHelper to prevent calls in the constructor to `getItemStackDamage()`.


MaPePeR:
Are you sure you even know what you are doing there? 

The damage value is needed to find the item in our emc map. (mapContains)

Robotia:
Ah, didn't notice that, I'll close this pull then. 😕

PR #1220, opened 2016-05-01
Fix unlearn condenser dupe

### LINKED ISSUE #1218: Free EMC forever.
Steps to replicate:

1. Open up a Transmutation Table/Tablet.
2. Place an item into the _unlearn_ slot.
3. Shift-click the item, turning it into EMC.  Notice a RED 0 in the lower-right corner of the item.
4. Left-click the item, picking it up. Place item into your inventory, manually. No shift-clicking or other assisted moves.
5. Manually place this RED0-Item into an Energy Condenser; left clicks and no manual placing. Receive the EMC again.

This starts a fairly easy EMC doubling loop.

MaPePeR:
Fixes #1218

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/container/TransmutationContainer.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/container/TransmutationContainer.java
@@ -86,11 +86,11 @@ public class TransmutationContainer extends Container
 		ItemStack stack = slot.getStack();
 		ItemStack newStack = stack.copy();
 		
-		if (slotIndex <= 7)
+		if (slotIndex <= 7) //Input Slots
 		{
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


MaPePeR:
This fix also probably needs to be backported to 1.7.10. I would be surprised, if the exploit would not exist in that version, too.

sinkillerj:
Mistakes were made, I kinda didint realize you were targeting MC18 with this, that being said it will be.

Robotia:
I can confirm it does exist in 1.7.10 :+1:

MaPePeR:
In retrospective it would have probably been wise to fix it in 1.7.10 and then just migrate the bugfix upwards with the other changes we are still doing in 1.7.10 - but too late for that, i guess :laughing:

Robotia:
TL;DR `Sorry not sorry`
:laughing:

PR #1223, opened 2016-05-04
Fix EMCProxyImpl.hasValue(ItemStack) ignoring Metadata. Fixes #1222

### LINKED ISSUE #1222: Issue with hasValue(ItemStack) API and items that use damage values
If you have a "multi-item" that uses damage values to represent multiple different items, the hasValue(ItemStack) API call can malfunction when the item with damage value 0 does not have an EMC value.

hasValue(ItemStack) appears to delegate to hasValue(stack.getItem()), which causes the item with damage value 0 to be checked instead of the actual item passed.

If the item passed has an EMC value, and the item with damage value 0 does not, the API call returns false.

This can easily be tested with the Thermal Foundation ingots. The item with damage 0 (Pulverized Iron Dust) has no EMC value, but other items with the same item ID but different damage values, like Copper Ingots, do have an EMC value. If you pass a stack of Copper Ingots to the API, it will return false. If you manually set an EMC value on the Pulverized Iron Dust, the API starts returning true when passed the stack of Copper Ingots.

MaPePeR:
This is only a GitHub-editor-commit. Did not check if it compiles or fixes anything... But it should.

### DIFF
--- a/src/main/java/moze_intel/projecte/api/proxy/IEMCProxy.java
+++ b/src/main/java/moze_intel/projecte/api/proxy/IEMCProxy.java
@@ -40,7 +40,7 @@ public interface IEMCProxy
     boolean hasValue(Block block);
 
     /**
-     * Queries the EMC value registry if the given item has an EMC value
+     * Queries the EMC value registry if the given item with a damage value of 0 has an EMC value
      * Can be called at any time, but will only return valid results if a world is loaded
      * Can be called on both sides
      * @param item The item we want to query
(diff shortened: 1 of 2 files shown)


williewillus:
IIRC the javadoc on the interface needs to be changed, as well, it states the calls are equivalent

MaPePeR:
Changed the JavaDoc

MaPePeR:
@sinkillerj this should have gone into the release. it fixes a critical bug in the API

PR #1226, opened 2016-05-08
Improvement upon [WIP] Finish ProjectE - Archangel homing arrows

MaPePeR:
Another idea to make it ark more would be to disallow the arrow changing the direction every tick.

Even with my changes the arrow can turn very fast, even though the max angle is restricted a lot. (I had problems that when restricting it even more it would be 0 all the time)

Try it out/test it and see if you like it.


blackdragonIV:
i would like to add a small note on the behavior of arrows when fighting the ender dragon, the majority of the fire arrows from the ring just keep homing on the top of the ender dragon trying to hit it but fail to do so. i think it's something related to the ender dragon hit box being on the head and the arrows trying to reach for its head but fail to do so and get stuck on its body, you can try it for yourself and you will get a more clear idea of what is happening.

sinkillerj:
After reviewing this pull currently the old code, albeit buggy, is better. This version fires beautifully but misses far too often for it to be useful. The problem could be alleviated if the arrow became more aggressive and went straight at the target if it knew it was going to hit the ground. NoClip would be a option as well, but underground mobs may be a problem.

williewillus:
just tweak the math so it curves more

sinkillerj:
Ill give it a shot (no pun intended).

sinkillerj:
Welp, Willie was right on the money with this one. And to think ive been dicking around with this thing for two hours, math is hard.

PR #1227, opened 2016-05-09
Allow Redmatter Morningstar to break RM/DM blocks

Robotia:
A player noticed a bug where RM and DM blocks are broken but not returned by the redmatter morningstar, I believe this change should fix the issue.


MaPePeR:
Please fix the indentation. Also: did you test this change?

Robotia:
Ok, I fixed the indentation, sorry about that! And I tested it, I am currently using it on PE1.7.10.

MaPePeR:
Thanks. Looks fine to me.

PR #1254, opened 2016-06-07
[1.9] Fix Wildcard in setValue.

### LINKED ISSUE #1245: Dark Oak Wood Has not EMC 1.9.4
Subj

MaPePeR:
Fixes #1245 

The OreDict entry for logWood uses Wildcard-values and i didn't account for that.


williewillus:
I wonder how it worked for the other logs then?

MaPePeR:
Its even more weird than that, because it worked for Acacia (Which shares the `minecraft:log2` with Dark Oak)

williewillus:
wait so do we know what causes the bug? + how's this fix it?

MaPePeR:
Yes. 

The OD defaults file contains "logWood" and sets a value for it. When the file is parsed it will call setValueBefore

PR #1269, opened 2016-06-17
[1.9] Fix harvestVein to properly check numMined for every recursion.

### LINKED ISSUE #1257: 1.9.4 dark matter pick right click ores issue
when i right click ores with dark matter  the ores disappear but the loot balls won't drop .the game lags and if you mine anything with the dark matter pick in that period you wouldn't get any drops.other tools like the dark matter shovel and vanilla picks work fine and drop blocks after doing the right click function 
the game will not crash and it gets  fixed after restarting the game

and after restarting the game the mined blocks aren't there (i mean it's not  desync or lag )

PS: the game freezes when you click exit but no crash log thats all  i got

MaPePeR:
Without this commit `numMined` will be reset, when one recursion ends and backtracks to the calling recursion again.

~~This is only a partial solution for #1257. It prevents the crash, but because for some reasons the `world.setBlockToAir()`method is not working it does mine the same block multiple times.~~

Found the problem. The block breaking code was not using the `currentPos` variable properly.

Fixes #1257

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
@@ -574,7 +574,7 @@ public abstract class PEToolBase extends ItemMode
 		for (BlockPos pos : WorldHelper.getPositionsFromBox(box))
 		{
 			IBlockState state = world.getBlockState(pos);
-			if (ItemHelper.isOre(state) && state.getBlockHardness(player.worldObj, pos) != -1 && (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, pos, stack)))
+			if (state != Blocks.AIR.getDefaultState() && ItemHelper.isOre(state) && state.getBlockHardness(player.worldObj, pos) != -1 && (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, pos, stack)))
 			{
 				WorldHelper.harvestVein(world, player, stack, pos, state, drops, 0);
 			}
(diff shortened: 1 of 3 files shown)


MaPePeR:
So.... Why did we do another 1.9 release without including this fix?

Bersham:
Not to be a dumbass, but I was wondering the same thing for the past couple weeks.

williewillus:
I haven't looked at this bc of time and @sinkillerj didn't inform me he was releasing or else I would've xP

sinkillerj:
It was just a quick hotfix update, its not like we cant just put out a new update at literately any time.

src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
```diff
@@ -574,7 +574,7 @@ protected void mineOreVeinsInAOE(ItemStack stack, EntityPlayer player) {
 		for (BlockPos pos : WorldHelper.getPositionsFromBox(box))
 		{
 			IBlockState state = world.getBlockState(pos);
-			if (ItemHelper.isOre(state) && state.getBlockHardness(player.worldObj, pos) != -1 && (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, pos, stack)))
+			if (state != Blocks.AIR.getDefaultState() && ItemHelper.isOre(state) && state.getBlockHardness(player.worldObj, pos) != -1 && (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, pos, stack)))
```
williewillus: whys this needed, doesn't the ore + hardness check take care of it?

MaPePeR: Don't know anymore. Might not be needed anymore, because I added the null check to the oredict function. 

The isOre function did crash for air blocks.

PR #1374, opened 2016-10-15
Update zh_CN.lang

3TUSK:
As #1324 has several fatal errors, I decided to make a new pull request instead.

Also... the StackOverflowError still persists with those translated manual entries. Looks like something still goes wrong when [loading manual entries](https://github.com/sinkillerj/ProjectE/blob/MC1.10.x/src/main/java/moze_intel/projecte/manual/ManualFontRenderer.java#L44). Afaik vanilla `FontRenderer` can handle it correctly...
Anyhow, I simply removed all manual entries, as requested by [previous commit](https://github.com/sinkillerj/ProjectE/commit/6d7c0ef56ac061e2fb10c16acc7f51e97c0d3c35).

If this pull request is merged, #1324 should be closed as well.


sinkillerj:
Thanks for this, I had my doubts about #1324 as well. Ill give this a look over and likely merge before the next build.

PR #1377, opened 2016-10-17
Fix crash when player inventory size is not 36

sjcl:
I got this crash report in 1.7.10-PE1.10.0. 
http://pastebin.com/q26ZwXrr

This, despite the Entity of LittleMaidMobX is inheriting the EntityPlayer, it is the cause the size of the inventory is 36 or less. As a result of this change, even if the size of the inventory is less than 36, it will not crash.


sjcl:
Thank you for changing the title. I'm sorry for my poor English.

sinkillerj:
ありがとう

PR #1409, opened 2016-11-21
Covalence Repair Changes

Lilylicious:
Instead of a "Correct recipe repairs full", it consumes EMC and repairs the equivalent damage.

### DIFF
--- a/src/main/java/moze_intel/projecte/utils/EMCHelper.java
+++ b/src/main/java/moze_intel/projecte/utils/EMCHelper.java
@@ -261,4 +261,21 @@ public final class EMCHelper
 		}
 		return 0;
 	}
+
+	public static int getEMCPerDurability(ItemStack stack){
+
+		int EMC;
+
+		if(stack == null)
+			return 0;
+
+		ItemStack stackCopy = stack.copy();
+		stackCopy.setItemDamage(0);
+
+		if(ItemHelper.isItemRepairable(stack)){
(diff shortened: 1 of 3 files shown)


src/main/java/moze_intel/projecte/utils/EMCHelper.java
```diff
@@ -261,4 +261,21 @@ private static double getStoredEMCBonus(ItemStack stack) {
 		}
 		return 0;
 	}
+
+	public static int getEMCPerDurability(ItemStack stack){
+
+		int EMC;
```
williewillus: `emc` instead of `EMC`

williewillus:
This diverges from EE2 behaviour but I think it's a good change to make

Lilylicious:
It does diverge, but it plugs up a pretty major exploit. It's a change that makes sense too, it makes it so that you can't get more EMC out of a repair than you put in. Also makes it easier to use, since repair recipes now don't care how much dust or what type you put in, as long as the EMC put in exceeds the minimum required to repair at all.

src/main/java/moze_intel/projecte/utils/EMCHelper.java
Lilylicious: Will include that in the final cleanup commit

williewillus:
@Lilylicious might want to remake this since the 1.10 branch got deleted (autocloses anything targeting it)

sinkillerj:
@williewillus Yea we already discussed it on IRC, Lily is gonna remake it after a few cleanup tweaks.

PR #1423, opened 2016-12-03
[MC17 Branch] Changed Int to Long

Wealthyturtle:
I Managed to Change Int to Long, Hope this Can be Applied in an Update.

EDIT: Need to Change Int to Long at More Locations


williewillus:
no

PR #1479, opened 2017-02-11
Mercurial Eye fleshed out and improved.

Stalkstalks:
I felt the need to get the mercurial eye up to speed, since it just never lived up to its old glory.

The mercurial eye now has the features I miss from the old EE2 versions, updated with modern quality-of-life improvements. I think it can now hold its own against other manual construction tool kits.

Extension, Transmutation, and Pillar Modes are implemented.

- Extension mode analyzes the block clicked on, and extends it and any adjacent identical blocks in the direction of the player. If the block doesn't have an EMC value, it will extend using the block stored in the eye. It's a lot like an EMC-powered builder's wand.
Charge levels allow it to extend 1/8/27/64/125 blocks at a time.
It can extend into transparent blocks such as tall grass or mushrooms without any problem - if they have EMC values, they will be accounted for in the cost of placing blocks; otherwise they will simply be broken and dropped if appropriate.

- Transmutation mode turns the block clicked on, as well as any adjacent identical blocks, into the stored block. It will use the EMC of the existing blocks to offset the cost, with any extra absorbed. If used to transmute blocks that don't have an EMC value it will break and drop them as appropriate, and constitute the replacement block from scratch.
Charge levels allow it to transmute 1/8/27/64/125 blocks at a time.

- Pillar mode creates or transmutes blocks to create large shafts of blocks. As with extension mode, it will use the block clicked on if it has an EMC value, and the stored block otherwise. Right clicking on empty air will place a single block of the stored type in mid-air (think angel blocks), which can then be extended into a full pillar with a second click.
Charge levels allow it to place 3x3x3/6/9/12/15 blocks at a time, just like classic EE2.

The eye now attempts to not place blocks where players or other entities currently occupy. Besides not being at risk for suffocating, it can be handy to use your body to reshape extensions.

Tapping left click will apply the current mode at minimum power to the block clicked. This is pretty useful with extension mode, where it can be used to quickly make breaks and turns in the blocks to be extended with a right click, marking out parts of the floor to raise into walls and then sculpting them to extend new floors and ceilings, etc. It's also pretty useful in the other modes mode. Try it out! This mercurial eye grants some pretty great blockbending powers. You can reshape landscapes into enormous structures quickly and easily.

I realize that 1.7 is kind of out of fashion these days but it's the version I'm still playing with my friends, so it's the one I worked on.

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/items/MercurialEye.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/items/MercurialEye.java
@@ -2,6 +2,7 @@ package moze_intel.projecte.gameObjs.items;
 
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
+import cpw.mods.fml.common.eventhandler.SubscribeEvent;
 import moze_intel.projecte.PECore;
 import moze_intel.projecte.api.item.IExtraFunction;
 import moze_intel.projecte.utils.Constants;
@@ -9,6 +10,8 @@ import moze_intel.projecte.utils.EMCHelper;
 import moze_intel.projecte.utils.PlayerHelper;
 import net.minecraft.block.Block;
 import net.minecraft.client.renderer.texture.IIconRegister;
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Blocks;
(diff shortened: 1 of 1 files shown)


williewillus:
Awesome work, you beat me to it :P
Only 1.10+ is active though, so I will probably pick your commits into a branch, make it compile, then merge into master after testing

Stalkstalks:
I hope it ports easily and that you and the users are happy with it.

If you like it, please consider including it on the main 1.7 branch so that it'd be easier to find again if I don't have my own copy when I'm on a nostalgia kick in 15 years. :p

williewillus:
one question, what is this kleinEmcCache field and how is it used?

Instance fields on Item/Block classes in Minecraft is a *huge* code smell because it is shared between ALL itemstacks of the item

Stalkstalks:
Ah, the code before me had variables storing the emc remaining in the klein star in method-scope variables and updating it every time the NBT tag on the klein star itself was updated.

I had thought that was quite strange myself, but assumed that the previous coder had some good reason for not just getting it from the item itself every time, probably performance related.

At one point my implementation grew to where I would be passing that value everywhere and it seemed like it would be better to keep it attached to the object instead of being referenced all over my methods and needing to be passed all over them despite being the same value for all of them. If you'd rather have it pulled as needed from whatever the new standard EMC storage scheme is, by all means go for that, I wasn't too psyched about doing it this way to begin with.

sinkillerj:
Ill consider it in the future but right now 1.7 is frozen. That said I'm with Willie on including it in 1.10+ once we give it a test run.

Stalkstalks:
Hey, williewillus, your mention of instance fields being shared across all stacks of the item made a bunch of things click about the minecraft codebase, and I had to amend my previous design decision even for my own little fork.

After getting a bit irritated that java doesn't let me pass doubles by reference, I saw that other mods don't shy away from just decoding the NBT tags for stored energy on every transaction, so I just went ahead and did that as well. Yeah, there's a performance hit, but it wasn't noticeable in testing, and if that's the accepted way to do it, it's at least not going to open the thing up to freakish behavior in the future.

williewillus:
why do add/remove emc return double now? looks like the value isn't being used.

Stalkstalks:
Ah, must have left that in while I was still trying to pass the value around instead of just calling it from the tag every time. Go ahead and turn it back to void if it has no use with the new EMC storage system.

williewillus:
@Stalkstalks I've put my attempt to port it on the `merceye` branch. Can you clone it and see if it all works as intended?

williewillus:
^ bump

williewillus:
#1824

PR #1513, opened 2017-05-27
Remove spaces

SupremeMortal:
Remove those spaces so the Linux commands can import the variables into the script, for instance, the source command in bash.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
-projecte_version = 1.7.10-PE1.10.1
-minecraft_version = 1.7.10
-forge_version = 10.13.2.1291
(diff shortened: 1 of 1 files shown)


sinkillerj:
As someone who compiles every single PE build on linux and has never had a problem... What are you on about here exactly?...

SupremeMortal:
I'm not talking about the compile process. I just wanted to read the variables from the gradle.properties in through the bash command `source` so I can use them as environment variables in my jenkins server. For some reason, the space in the file cause the `source` command to read them in as commands and not variables.

williewillus:
Wrong target branch either way :P

PR #1568, opened 2017-08-15
Create zh_tw.lang

CLOVERsquare:
The manual and some advancement are not translated yet.


williewillus:
Leave untranslated keys out of the file instead of copying. The values will be inherited from en_US automatically

PR #1634, opened 2018-01-04
IItemEmc fixes

### LINKED ISSUE #1632: IItemEmc assumptions
IItemEmc assumes klein star in many cases

Mordenkainen:
Fixes for various issues where PE made some assumptions when dealing with IItemEmcs and such.

There may be other locations I haven't found yet.

fixes #1632

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
@@ -2,6 +2,7 @@ package moze_intel.projecte.gameObjs.container.slots.transmutation;
 
 import moze_intel.projecte.api.item.IItemEmc;
 import moze_intel.projecte.gameObjs.container.inventory.TransmutationInventory;
+import moze_intel.projecte.gameObjs.container.slots.SlotPredicates;
 import moze_intel.projecte.utils.EMCHelper;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.items.SlotItemHandler;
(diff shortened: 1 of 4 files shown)


src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
```diff
@@ -51,7 +52,9 @@ public void putStack(@Nonnull ItemStack stack)
 			}
 		}
 
-		inv.handleKnowledge(stack.copy());
+		if (EMCHelper.doesItemHaveEmc(stack)) {
+			inv.handleKnowledge(stack.copy());
```
williewillus: what's the if needed for, doesn't the isItemValid change guarantee anything coming in to putStack have a emc value?

src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotLock.java
```diff
@@ -53,7 +54,9 @@ public void putStack(@Nonnull ItemStack stack)
 			}
 		}
 		
-		inv.handleKnowledge(stack.copy());
+		if (EMCHelper.doesItemHaveEmc(stack)) {
+			inv.handleKnowledge(stack.copy());
+		}
```
williewillus: same here

src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
Mordenkainen: actually no.... It is an assumption that because something is an IItemEmc that it also has an EMC value, but it doesn't have to. For example, my EMC storage cells are IItemEmc's, but have no EMC value of their own.

src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotLock.java
Mordenkainen: same here

PR #1649, opened 2018-02-18
Added en_UD.lang

The-Fireplace:


sinkillerj:
In order to accept this PR I'm going to need some more info on why you find it to be useful. If this is just one of Minecrafts joke languages its not worth implementing and maintaining.

The-Fireplace:
Yeah, it's just one of Minecraft's joke languages. I might make a gradle
plugin at some point to auto-maintain it.

sinkillerj:
Should that happen I would be interested.

The-Fireplace:
I'll make a new PR when I have a gradle plugin to auto-generate it.

PR #1650, opened 2018-02-18
Fix a few bugs

TeamSpen210:
* Check for NBT, not item damage for Void Rings and Black Hole Band in bags. This allows them to suck items into the bags again.
* Only tick down the Void Ring teleport cooldown when positive. Otherwise it goes negative, and wraps around constantly.
* Fix a wrong null annotation causing a crash on startup. (Color should be null to sync all of them, and the functions handle this correctly.)


williewillus:
Thanks!

PR #1651, opened 2018-02-18
Backport fix for issue #1632

Mordenkainen:
Backport of fix for issue #1632


williewillus:
We're not making new releases for older versions as far as I know (sin can reopen if he feels otherwise, but this is 1.7...)

forresthopkinsa:
1.7.10 is still very widely used, I think your stance on backporting should be revisited.

PR #1698, opened 2018-06-05
Longs for EMC. And bug fixed.

VladoCC:
All emc related numbers now stored in Longs. Fixed bug with Novas (now you can break them without crash).


MaPePeR:
Why delete all the tests?

VladoCC:
Sorry. Do not have much time to rewrite it o new arithmetics and forget about them.

VladoCC:
Now with your tests

VladoCC:
And, so, this will be merged?

MaPePeR:
This will get the process rolling, but i assume it will need extensive considerations and review.

It cannot be just merged as is.

src/main/java/moze_intel/projecte/emc/SimpleGraphMapper.java
```diff
@@ -186,7 +186,7 @@ private V valueForConversionUnsafe(Map<T, V> values, Conversion conversion)
 					continue;
 				}
 				//value = value + amount * ingredientcost
-				V ingredientValue = conversion.arithmeticForConversion.mul(entry.getValue(),values.get(entry.getKey()));
+				V ingredientValue = conversion.arithmeticForConversion.mul(entry.getValue().intValue(),values.get(entry.getKey()));
```
williewillus: why is this not longValue?

src/main/java/moze_intel/projecte/emc/mappers/OreDictionaryMapper.java
```diff
@@ -38,8 +38,8 @@ public void addMappings(IMappingCollector<NormalizedSimpleStack, Integer> mapper
 							continue;
 						}
 
-						mapper.setValueBefore(NSSItem.create(stack), 0);
-						mapper.setValueAfter(NSSItem.create(stack), 0);
+						mapper.setValueBefore(NSSItem.create(stack), 0l);
+						mapper.setValueAfter(NSSItem.create(stack), 0l);
```
williewillus: explain?

src/main/java/moze_intel/projecte/gameObjs/container/CondenserContainer.java
```diff
@@ -93,10 +93,21 @@ public void detectAndSendChanges()
...
+		if (id % 10 == 0){
+			if (id == 0){
+				displayEmc = 0;
+			}
+			displayEmc += data;
+		} else if (id % 10 == 1){
+			if (id == 1){
+				requiredEmc = 0;
+			}
+			requiredEmc += data;
```
williewillus: uh what

src/main/java/moze_intel/projecte/gameObjs/container/CondenserMK2Container.java
```diff
@@ -58,7 +58,7 @@ public ItemStack transferStackInSlot(EntityPlayer player, int slotIndex)
 			return ItemStack.EMPTY;
 		}
 
-		Slot slot = this.getSlot(slotIndex);
+;		Slot slot = this.getSlot(slotIndex);
```
williewillus: unnecessary

src/main/java/moze_intel/projecte/gameObjs/items/tools/PEToolBase.java
```diff
@@ -292,8 +292,8 @@ protected void digBasedOnMode(ItemStack stack, World world, Block block, BlockPo
 
 			if (b != Blocks.AIR
 					&& state.getBlockHardness(world, digPos) != -1
-					&& (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, digPos, stack))
-					&& PlayerHelper.hasBreakPermission(((EntityPlayerMP) player), digPos))
+					&& PlayerHelper.hasBreakPermission(((EntityPlayerMP) player), digPos)
+					&& (canHarvestBlock(state, stack) || ForgeHooks.canToolHarvestBlock(world, digPos, stack)))
```
williewillus: explain?

src/main/java/moze_intel/projecte/network/PacketHandler.java
```diff
@@ -71,15 +71,15 @@ public static void sendFragmentedEmcPacketToAll()
...
-		int[][] ret = new int[EMCMapper.emc.size()][];
+		long[][] ret = new long[EMCMapper.emc.size()][];
 		int i = 0;
-		for (Map.Entry<SimpleStack, Integer> entry : EMCMapper.emc.entrySet())
+		for (Map.Entry<SimpleStack, Long> entry : EMCMapper.emc.entrySet())
 		{
 			SimpleStack stack = entry.getKey();
 			int id = Item.REGISTRY.getIDForObject(Item.REGISTRY.getObject(stack.id));
-			ret[i] = new int[] { id, stack.damage, entry.getValue() };
+			ret[i] = new long[] { id, stack.damage, entry.getValue() };
```
williewillus: I'm in favor of making this an actual object type so we don't have to waste space on the IDs, but I'll let it slide and fix it after merging this

src/main/java/moze_intel/projecte/network/commands/SetEmcCMD.java
```diff
@@ -59,7 +59,7 @@ public void execute(@Nonnull MinecraftServer server, @Nonnull ICommandSender sen
 
 			name = heldItem.getItem().getRegistryName().toString();
 			meta = heldItem.getItemDamage();
-			emc = MathUtils.parseInteger(params[0]);
+			emc = Long.parseLong(params[0]); //MathUtils.parseInteger(params[0]);
```
williewillus: remove the old commented code

src/main/java/moze_intel/projecte/network/commands/SetEmcCMD.java
```diff
@@ -92,7 +92,7 @@ public void execute(@Nonnull MinecraftServer server, @Nonnull ICommandSender sen
 				}
 				else
 				{
-					emc = MathUtils.parseInteger(params[1]);
+						emc = MathUtils.parseInteger(params[1]);
```
williewillus: whitespace

src/main/java/moze_intel/projecte/network/packets/UpdateWindowLongPKT.java
```diff
@@ -9,16 +9,16 @@
 import net.minecraftforge.fml.common.network.simpleimpl.MessageContext;
 
 // Version of SPacketWindowProperty that does not truncate the `value` arg to a short
-public class UpdateWindowIntPKT implements IMessage
+public class UpdateWindowLongPKT implements IMessage
```
williewillus: don't rename this. make a new one for long and only use it when necessary

I don't want to waste 4 more bytes per update for fields that don't need it

src/main/java/moze_intel/projecte/utils/PlayerHelper.java
```diff
@@ -134,7 +134,7 @@ public static BlockPos getBlockLookingAt(EntityPlayer player, double maxDistance
 	public static boolean hasBreakPermission(EntityPlayerMP player, BlockPos pos)
 	{
 		return hasEditPermission(player, pos)
-				&& ForgeHooks.onBlockBreakEvent(player.getEntityWorld(), player.interactionManager.getGameType(), player, pos) != -1;
+				&& !(ForgeHooks.onBlockBreakEvent(player.getEntityWorld(), player.interactionManager.getGameType(), player, pos) == -1);
```
williewillus: explain?

williewillus:
1. Remove everything that isn't directly related to the EMC change (null annotation crash, breaking checks, etc.). The null annotation crash can be "fixed" by disabling "runtime checks for Nullability annotations" in your IDE.
2. Fix the tests so they compile, but don't change any of the existing test cases (it seems like you just appended a "1" to everything???....). Make a new test class if you want to exercise large values.

src/main/java/moze_intel/projecte/gameObjs/items/rings/ArchangelSmite.java
```diff
@@ -132,7 +132,7 @@ public void updateInPedestal(@Nonnull World world, @Nonnull BlockPos pos)
 			{
 				if (!world.getEntitiesWithinAABB(EntityLiving.class, tile.getEffectBounds()).isEmpty())
 				{
-					for (int i = 0; i < 3; i++)
+					for (int i = 0; i < 20; i++)
 					{
```
MaPePeR: This makes the Archangel more powerful in a pedestal?

src/main/java/moze_intel/projecte/utils/PlayerHelper.java
VladoCC: Can not remember why, but this works better.

Remove? Hm.. It changes something? No. Why do not just use this version?

VladoCC:
You trying to criticize small issues in my code and ignore your own.
If you are so criticizing, then just start to read code that already commited to this repo (I did it and know what I am talking about) and you will need to rewrite every second class.

src/main/java/moze_intel/projecte/utils/PlayerHelper.java
MaPePeR: Because you cannot provide an explanation for why to use your version or why your version is better? Also this line probably doesn't change a single thing....

`!(a == -1)` <=> `(a != -1)` ...

src/main/java/moze_intel/projecte/gameObjs/items/rings/ArchangelSmite.java
MaPePeR: Its not good practice to sneak changes like this in with a big PR like this, without announcing them.

williewillus:
Last warning, please remove all changes unrelated to EMC longs. 

Regarding nitpicking, the codebase can be messy, while is exactly why we should hold PR code to a *higher* standard not a lower one. My comments are how any review will look like.

VladoCC:
Ok, I am ready to remove all unrelated changes, but LongPkt will be still used everytime
4 bytes is nothing nowadays.

And @williewillus test are correct.
This is not "1",this is "l" (lower cased "L") to say for Java that we will work with Long. Same as "f" for Float.

williewillus:
4 bytes *is* something when every mod thinks like that, which is how you get [100kb/s bandwidth consumption](https://github.com/TerminatorNL/LagGoggles/issues/23#issuecomment-375262532) when standing in a modded base

as for the tests, there's no need to add the L then, ints are auto-promoted to long

pupnewfster:
As it seems that VladoCC is not really interested in bothering to make this pull up to spec (which I believe is because he mainly did some edits for his personal use) I am going to fork his repo and deal with the changes requested.

> You can change what you want in this pull.
> I just gave you example of how it needs to work.


I find the concept of being able to use longs for emc (finally) to be very appealing. And would like to see it finally happen. As I will be forking it the changes he has made will still be referencing him in the commit history so he will still get credit for his work.

One question I do have about the byte difference is I will make it so the objects that are always integers use an integer packet; however I am wondering if it is also wanted to make it so that if the emc value can fit inside an int that it also just uses an int packet or can I just have emc values always be in a long. (It was a bit unclear if you were just talking about the other variables sent or all of them).

I also will fix a few bugs that I noticed in places with how the long <-> int conversion is handled. And remove all changes not pertaining to long version.

MaPePeR:
I would prefer the EMC generation to continue to use Rationals/Fractions instead of Doubles like in the PR. Don't know if there is a Long version of Fraction, though.

williewillus:
re: the long/integer packets

I just meant if the data type being sent is an int, then send the int packet, otherwise I'm totally fine with sending long packets (no need to do weird compression stuff)

VladoCC:
@pupnewfster good luck)
Longs will make ProjectE better

pupnewfster:
@MaPePeR Looking at it I was unable to find something like Rationals/Fractions for Longs. However, upon trying to fix the last test (GraphMapperTest.testOverflowWithAmount) that is erroring on me (I still have not dealt with some of the other errors I have seen)  I found something that may work. The issue with this test is that double rounding math is not all that accurate
`java.lang.AssertionError: expected:<4611686018427387903> but was:<4611686018427387904>` so upon looking for something more accurate I ran into [BigDecimal](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html) which seems to have some methods similar to Fraction.

If you think this is close enough then I will attempt to convert the use of Doubles to using BigDecimal so that the precision is not lost during calculations.

MaPePeR:
There is also [BigFraction](http://commons.apache.org/proper/commons-math/javadocs/api-3.3/org/apache/commons/math3/fraction/BigFraction.html), which might be a better replacement. If you look at the code before the changes in this PR there was a Fraction-Arithmetic. Maybe its as easy as adding "Big" in front of the Type name.

pupnewfster:
There is one problem with that it is in apache commons-math which is not included in forge. Should I attempt to import the package (which would end up making the jar a lot larger as I think it would have to include apache commons-math so that it can use it)? OR should I attempt to use BigDecimal to fix the Double accuracy and maybe at some point later down the line come up with a better solution.

MaPePeR:
Oh. I was under the impression that that library is already included in minecraft. I'm sure some of the apache commons libraries are. Apparently not math?

Thanks for testing this out.

MaPePeR:
[BigFraction](https://commons.apache.org/proper/commons-math/javadocs/api-3.6/org/apache/commons/math3/fraction/BigFraction.html) is part of the same library we [were already using](https://github.com/sinkillerj/ProjectE/blob/mc1.12.x/src/main/java/moze_intel/projecte/emc/arithmetics/HiddenFractionArithmetic.java#L3).

Not sure, what changes between Fraction and BigFraction that causes the loop-recipes to error out, though.

pupnewfster:
I have figured out why it was causing them to time out. It is because BigFraction (like other Big types) goes to further accuracy than a long can be. so when it is getting infinitely closer to zero it times out. I managed to fix this by adding a check to see if the denominator ends up overflowing a long because it does not end up throwing the arithmeticexception like before.

I will see what is involved about potentially having it include the BigFraction stuff so that then it can be used keeping it like how it was before with Fractions.

pupnewfster:
Ok @MaPePeR I have managed to get it working with BigFraction and it seems to work perfectly (both in the tests and in game). The one down side is I had to shadow the jar to include commons.math3 so the jar's file size went from 1,453 KB to 3,742 KB. If you would like me to commit this change to my updated version of this pull, #1709 I can. One thing that should be looked at is how I do the shadow jar given I am not sure if it messes with any of the custom gradle stuff prjecte uses as I am still relatively new to gradle.

If you guys do not want the increase in file size I can attempt to see if I can get BigDecimal working properly and then use that as it is built in to java. Though its formatting probably will be closer to doubles than fractions.

zeebrandonite:
There are quite a few items that can't be given an EMC value because int is simply too small, so I humbly request that the EMC values in ProjectE version 1.7.10-PE1.10.1 be converted into longs, it would be a lifesaver. 
P.S. I have been looking for something like this for at least 5 years.

Issue #1706, opened 2018-06-11
Request for two new Events


pupnewfster:
Exact ProjectE version (do not say "latest", "latest on Curse", or similar):  `1.12-PE1.3.0`

Exact Forge version: `forge-14.23.4.2707`

No crash or steps to reproduce as this is a feature/api request. It does not quite follow the formatting but I will attempt to format the below request in a readable manner.

## Preface

I am one of the devs for the mods Reskillable and CompatSkills which allow for pack creators to lock/gate a large number of aspects of the game and of other mods.

## The request

I would appreciate if two cancelable events could be added to ProjectE:

1. An event that allows cancelling "learning" a new item in the transmutation table. For my purposes it does not matter if it stops the person from using/gaining the emc from the item but if it could be able to stop them from "learning" the item and being able to produce more of it that would be great.
2. An event that allows cancelling putting an item in the "tramutation" slot of energy condensers.

Both of these events would need to have the player object attempting to perform the task, and the item stack they are trying to use. This would be so that I can ensure the player meets the requirements.

## Use Cases

The goal of these events would be to allow me to add support for staging ProjectE without just requiring harder recipes and then once you finally get to a transmutation table you can just keep on gathering raw materials/emc to progress further.

With these events I would be able to add handlers to CompatSkills to allow locking the ability to "learn" and "transmute" items into other items if they cost above a certain amount of emc. For example: I could lock the ability to transmute diamonds until specific requirements are met while still allowing players to transmute iron ingots into food.

pupnewfster:
I have decided to work on these myself and I will create a pull request sometime this week (maybe tonight if I manage to get it working).

PR #1707, opened 2018-06-13
Support cancelling learning and setting condenser duplicator item

pupnewfster:
Implements #1706 

The general description can be found in the issue above. For the implementation of the AttemptLearnEvent I decided that cancelling transmutation learning would still give the player emc; however, it does not add the item to their "learned" items.

EDIT: One thing I forgot to mention is I have fully tested the code and it works properly. If needed I can push my local repo changes to CompatSkills to a branch on my fork of it so that it is easier to test for you guys.

### DIFF
--- a/src/main/java/moze_intel/projecte/api/event/PlayerAttemptCondenserSetEvent.java
+++ b/src/main/java/moze_intel/projecte/api/event/PlayerAttemptCondenserSetEvent.java
@@ -1,0 +1,44 @@
+package moze_intel.projecte.api.event;
+
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fml.common.eventhandler.Cancelable;
+import net.minecraftforge.fml.common.eventhandler.Event;
+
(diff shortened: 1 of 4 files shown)


pupnewfster:
I am linking two of the images I took from when I was testing that I realized may be of use in understanding the use case of this pull.

https://i.gyazo.com/88721f03cfb8de49a7e0b1aa48957646.png
https://i.gyazo.com/a3c18415a3fd28cceec65bb24bf64860.png

The specific requirements are just what I had set to diamond for testing, however the `Magic: 10` requirement I have set for all items that cost 8192 emc or higher.

williewillus:
Looks good, just document what sides the event is fired on

pupnewfster:
Ok, at the moment at least it fires on both the server and client because I was running into strange sync issues with the gui's when I did it on just the serveror the client. Would you like me to write that it fires on both, or just leave it as is?

pupnewfster:
Ok I modified it to say that it fires it both server side and client side. The issue is that if it does not get fired on both then it causes desyncs with the gui where the player thinks they have placed it but have not etc.

PR #1709, opened 2018-06-18
Long values for EMC

pupnewfster:
Continuation/fix of #1698 

Comments from original pull that got addressed:
- Only sends the `UpdateWindowLongPKT` for EMC instead of all update window packets
- Removed all changes not pertaining to the switch to longs
- Fixed a bunch of bugs I found in the code. The only remaining one is the double accuracy bug listed below.
- Switched from `l` to `L` for long literals as unlike @williewillus was saying it cannot autopromote integers to Long objects, only to long primitives.

It currently still uses doubles which is the only point that has not been addressed yet. This does cause accuracy issues for very large longs due to double accuracy and I was discussing with @MaPePeR  about switching to BigFraction or BigDecimal. More information can be found in [this comment](https://github.com/sinkillerj/ProjectE/pull/1698#issuecomment-397891036) and the ones following it.

I decided to open this pull so the other changes I have made can be looked at and if people want to test they can.

Edit: The switch to BigFraction has been made.


src/main/java/moze_intel/projecte/network/PacketHandler.java
```diff
@@ -71,15 +80,15 @@ public static void sendFragmentedEmcPacketToAll()
...
-		for (Map.Entry<SimpleStack, Integer> entry : EMCMapper.emc.entrySet())
+		for (Map.Entry<SimpleStack, Long> entry : EMCMapper.emc.entrySet())
 		{
 			SimpleStack stack = entry.getKey();
 			int id = Item.REGISTRY.getIDForObject(Item.REGISTRY.getObject(stack.id));
-			ret[i] = new int[] { id, stack.damage, entry.getValue() };
+			ret[i] = new long[] { id, stack.damage, entry.getValue() };
 			i++;
 		}
 		PECore.debugLog("EMC data size: {} bytes", ret.length * 3 * 4);
```
williewillus: update this to `* 3 * 8` since longs are 8 bytes long

src/main/java/moze_intel/projecte/network/packets/SyncEmcPKT.java
```diff
@@ -30,15 +30,21 @@ public SyncEmcPKT(int[][] data)
...
-			int[] array = new int[3];
+			long[] array = new long[3];
 
 			for (int j = 0; j < 3; j++)
 			{
-				array[j] = ByteBufUtils.readVarInt(buf, 5);
+				int val = ByteBufUtils.readVarInt(buf, 5);
+				if (j == 2) { //EMC it is stored in this position and the next one
+					//Pack the two ints representing the bits of the EMC long back into a long to retrieve the true value
+					array[j] = (long) val << 32 | ByteBufUtils.readVarInt(buf, 5) & 0xffffffffL;
```
williewillus: does ByteBuf and ByteBufUtils really not have methods for longs? that's surprising

src/main/java/moze_intel/projecte/network/packets/SyncEmcPKT.java
```diff
@@ -17,11 +17,11 @@
 
 public class SyncEmcPKT implements IMessage
 {
-	private int[][] data;
+	private long[][] data;
```
williewillus: if you feel like it, now would be a good time to make this strongly typed (a class with `int id`, `int damage`, `long data`) or whatever, instead of an ugly 2d array. not a big deal though.

pupnewfster:
I am working on cleaning up how I shade the jar and I will finish when I get back in a few hours and then push.

MaPePeR:
Looks good to me. One thing i'm thinking about if its possbile to use the shadowJar thing to only include the classes we actually use to reduce the jar size. Does someone know if it that is possible? I think i saw some `include` and `exclude` directives in the documentation, which we might be able to use?

MaPePeR:
PR passed my smoke test. Looks good to me. 👍

sinkillerj:
Looks good here. Ill leave it to willie to make the final call on merge, will put out a test build once thats done.

williewillus:
welp we'll see how this goes

PR #1714, opened 2018-06-27
CraftTweaker Recipe Mapping

pupnewfster:
The new CraftTweaker mapper is only used for newer versions of CraftTweaker that have their new recipe system.

This preserves compatibility with older versions of CraftTweaker

### DIFF
--- a/src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java
+++ b/src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java
@@ -1,24 +1,17 @@
 package moze_intel.projecte.emc.mappers;
 
-import com.google.common.collect.Iterables;
+import crafttweaker.mc1120.recipes.MCRecipeShaped;
+import crafttweaker.mc1120.recipes.MCRecipeShapeless;
 import net.minecraft.item.crafting.IRecipe;
 import net.minecraftforge.fml.common.Loader;
-import net.minecraftforge.fml.common.ModAPIManager;
-import net.minecraftforge.fml.common.ModContainer;
 import net.minecraftforge.fml.common.versioning.VersionParser;
 
 public class CraftTweakerRecipeMapper implements CraftingMapper.IRecipeMapper {
     private boolean ctCompat;
 
     public CraftTweakerRecipeMapper() {
-        if (Loader.isModLoaded("crafttweaker")) { //Check to make sure it is a version of CraftTweaker that uses the new Recipe System
-            for (ModContainer mod : Iterables.concat(Loader.instance().getActiveModList(), ModAPIManager.INSTANCE.getAPIList())) {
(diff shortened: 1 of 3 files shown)


src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraftforge.fml.common.ModAPIManager;
+import net.minecraftforge.fml.common.ModContainer;
+import net.minecraftforge.fml.common.versioning.VersionParser;
+
+public class CraftTweakerRecipeMapper implements CraftingMapper.IRecipeMapper {
+    private boolean ctCompat;
+
+    public CraftTweakerRecipeMapper() {
+        if (Loader.isModLoaded("crafttweaker")) { //Check to make sure it is a version of CraftTweaker that uses the new Recipe System
+            for (ModContainer mod : Iterables.concat(Loader.instance().getActiveModList(), ModAPIManager.INSTANCE.getAPIList())) {
```
williewillus: is this necessary? can't you just get the modcontainer from `Loader`?

williewillus:
Also, I don't understand how this resolves the issue - are the CT recipe classes subclasses of vanilla classes? If so, the earlier mappers in the list will load and process it. If not, then how is the mapper providing info about the ingredients to the calculation system?

src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java
pupnewfster: From what I could find by looking at the Loader class there is no method to retrieve ModContainer directly from the Loader.

src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Maps `IRecipe` CraftTweaker crafting recipes that extend `MCRecipeShaped` or `MCRecipeShapeless";
+    }
+
+    @Override
+    public boolean canHandle(IRecipe recipe) {
+        //Make sure no imports are added that could cause problems
```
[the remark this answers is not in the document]
pupnewfster: Basically it is to ensure that it doesn't throw a class not found exception because the mapper gets registered even if the classes do not exist. And @MaPePeR also has found that in the past sometimes imports cause issues even though the majority of times they don't.

pupnewfster:
And it provides info about the ingredients to the calculation system because it implements IRecipe, the two recipe types are just not instances of the vanilla classes due to how they structure things internally. But given they are shaped/shapeless it only needs to get the ingredient list (which IRecipe includes a method to do so).

PR #1722, opened 2018-07-06
Klein Star and JEI fixes

Lilylicious:
Klein stars now transfer EMC when upgrade crafting

Klein stars show up in JEI

JEI has been modernized to 1.12 standards

JEI GUI for world transmutation has been fixed


Lilylicious:
The latest commit added ditches the custom container for the philosopher's stone crafting grid, which had no additional functionality compared to the default container that JEI recognizes.

The only difference I found was the method canInteractWith. This method is not run as part of the container creation/use, it seems to be purely a pre-use check. As such, it is not necessary to return true for the philosopher's stone to use the GUI and crafting grid.

src/main/java/moze_intel/projecte/utils/GuiHandler.java
```diff
@@ -129,7 +128,7 @@ public Object getServerGuiElement(int ID, EntityPlayer player, World world, int
 			case Constants.MERCURIAL_GUI:
 				return new MercurialEyeContainer(player.inventory, new MercurialEyeInventory(player.getHeldItem(hand)));
 			case Constants.PHILOS_STONE_GUI:
-				return new PhilosStoneContainer(player.inventory);
+				break;
```
williewillus: just remove this constant and the `case`

Lilylicious: I didn’t originally, but I opted for keeping them to preserve the number order and to make it easier to add something new in the future. Removing them is simple enough though.

Lilylicious:
Just noticed issues with some world transmutations. Not ready to be merged yet.

Lilylicious:
All fixed!

Lilylicious:
GUI changes turned out to be bad and shouldn't have been done. Transferhandler added instead.

Lilylicious:
Ready for a merge unless you can throw more JEI related things we're missing at me.

PR #1724, opened 2018-07-09
Convert tile entities and items to fully use Long

Tencao:
This is not final yet as I'm still not happy with the state of the rings and they'll probably need changing, but making the PR so I can gather feedback in the meantime.


williewillus:
don't mix in non-long/double related changes (like the timer stuff)

Tencao:
We was discussing this on discord yesterday, this was accepted as a way to avoid changing numbers of generation while still providing the same result. The only one this is mostly an issue for was item costs as there's currently no easy way of changing them without flat out increasing or removing their cost

MaPePeR:
The general idea was to get away from double and also use long for storage.

This would need conversion code for backwards compatibility, though. Wouldn't it?

Lilylicious:
To sum up the reasoning behind the timers and balancing: Lots of the EMC cost/balancing was done with fractions, thus the doubles. If we no longer use doubles, we cannot keep the same values as before.

Tencao:
Basically what @Lilylicious said, the other solution to not using timers is increasing or removing emc cost, and that's not something I wanted to decide on my own, so I uploaded what I've done to get solutions moving forward.

gradlew
```diff
@@ -114,6 +109,7 @@ fi
 if $cygwin ; then
     APP_HOME=`cygpath --path --mixed "$APP_HOME"`
     CLASSPATH=`cygpath --path --mixed "$CLASSPATH"`
+    JAVACMD=`cygpath --unix "$JAVACMD"`
```
MaPePeR: Why this change? This has nothing to do with the Long-thing

src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
```diff
@@ -38,7 +38,7 @@ public void putStack(@Nonnull ItemStack stack)
 		if (stack.getItem() instanceof IItemEmc)
 		{
 			IItemEmc itemEmc = ((IItemEmc) stack.getItem());
-			double remainingEmc = itemEmc.getMaximumEmc(stack) - (int) Math.ceil(itemEmc.getStoredEmc(stack));
+			long remainingEmc = itemEmc.getMaximumEmc(stack) - (int) Math.ceil(itemEmc.getStoredEmc(stack));
```
MaPePeR: casting to int here feels wrong?

src/main/java/moze_intel/projecte/gameObjs/items/rings/HarvestGoddess.java
```diff
@@ -52,25 +53,21 @@ public void onUpdate(ItemStack stack, World world, Entity entity, int par4, bool
...
+		if (player.getCapability(InternalTimers.CAPABILITY, null).canRingUpdate()) {
+
+			if (ItemHelper.getOrCreateCompound(stack).getBoolean(TAG_ACTIVE)) {
+				long storedEmc = getEmc(stack);
+
+				if (storedEmc == 0 && !consumeFuel(player, stack, 64, true)) {
+					stack.getTagCompound().setBoolean(TAG_ACTIVE, false);
+				} else {
+					WorldHelper.growNearbyRandomly(true, world, new BlockPos(player), player);
+					removeEmc(stack, 6);
```
MaPePeR: 6 emc every n ticks does not add up to 0.32 emc per tick?

src/main/java/moze_intel/projecte/gameObjs/items/rings/SWRG.java
```diff
@@ -101,23 +102,21 @@ private void tick(ItemStack stack, EntityPlayer player)
...
-			toRemove = 0.32F;
-		}
-		else if (stack.getTagCompound().getInteger(TAG_MODE) == 3)
-		{
-			toRemove = 0.64F;
-		}
+			if (stack.getTagCompound().getInteger(TAG_MODE) == 2) {
+				toRemove = 6;
+			} else if (stack.getTagCompound().getInteger(TAG_MODE) == 3) {
+				toRemove = 12;
```
MaPePeR: 6 and 12 do not add up to 0.32 or 0.64

src/main/java/moze_intel/projecte/gameObjs/tiles/CollectorMK1Tile.java
```diff
@@ -171,76 +173,75 @@ else if (!getUpgrading().isEmpty())
...
-			this.addEMC(getSunRelativeEmc(emcGen) / 20.0f);
+	private void updateEmc() {
+		if (emcTimer > 0) {
+			emcTimer--;
+		} else {
+			if (!this.hasMaxedEmc()) {
+				this.addEMC(getSunRelativeEmc(emcGen) / 4);
+			}
+			emcTimer = 5;
 		}
```
MaPePeR: This file has a lot of formatting changes that make reviewing hard

src/main/java/moze_intel/projecte/utils/WorldHelper.java
```diff
@@ -510,7 +510,7 @@ public static void igniteNearby(World world, EntityPlayer player)
 	{
 		for (BlockPos pos : BlockPos.getAllInBoxMutable(new BlockPos(player).add(-8, -5, -8), new BlockPos(player).add(8, 5, 8)))
 		{
-			if (world.rand.nextInt(128) == 0 && world.isAirBlock(pos))
+			if (world.rand.nextInt(32) == 0 && world.isAirBlock(pos))
```
MaPePeR: This is a balance change?

gradlew
Tencao: This was an automatic change, I haven't touched that and can be freely discarded

src/main/java/moze_intel/projecte/gameObjs/container/slots/transmutation/SlotInput.java
Tencao: Yeah it does, I'll probably change that to long

src/main/java/moze_intel/projecte/gameObjs/items/rings/HarvestGoddess.java
Tencao: Was one of the many examples of how it doesn't translate well. I need a better solution to all items, not just this one

src/main/java/moze_intel/projecte/utils/WorldHelper.java
Tencao: It was only to compensate for the less ticks it'll be doing, if we find a better solution this could be reverted

src/main/java/moze_intel/projecte/gameObjs/tiles/CollectorMK1Tile.java
Tencao: I tried to only attach the timers to parts actually generating emc so the rest of the tile can function as normal

src/main/java/moze_intel/projecte/gameObjs/items/rings/SWRG.java
Tencao: Items are not final and need a better solution, this was only temp and done for the sake of testing

pupnewfster:
Thinking about it, one of the better solutions might be to have the balance be slightly different from what it is currently and add config options for changing rates of things. That way pack makers can change things if they want to. And just default it to the closest value it was before.

PR #1735, opened 2018-08-11
Fix jei [+] for philosopher's stone

### LINKED ISSUE #1703: PStone crafting table missing the recipe + button
ProjectE: 1.12.2 PE 1.3.0

Forge: 14.23.3.2691

Steps to reproduce:
Use C to open PStone crafting table.
Examine recipe with JEI (1.12.2-4.9.1.179).
No + button to move recipe to crafting table.

+ button is usable on basic crafting table.

malakai97:
Resolves #1703

I don't really understand this behavior, but it is recommended by
JEI's wiki and it resolves the issue.


pupnewfster:
The reason that it seems broken normally is the merged code that fixes it is not in a released version yet. It was fixed in #1722

Lilylicious:
Can confirm, this change does absolutely nothing. The fix is already in the code, but is not yet present in a release version of ProjectE. If you revert this change and run it, it still works.

PR #1771, opened 2018-11-26
Fix for server-side crash from JEI

Lilylicious:
I'm really really hoping this'll fix it. I can't test, due to not getting the crash when running a client and server on my local machine.

getMinecraftServerInstance() is null when running in certain client/server configurations. Minecraft.getMinecraft() should always exist when a client is running it, which we're checking for.

On top of this, I added a fix for a bug that I noticed while debugging the feature, where the final item in the fuel map would display a recipe going from finalItem -> firstItem due to the wrapping effects of getFuelUpgrade.


williewillus:
I can give this a test today

PR #1780, opened 2018-12-02
Added abreviations for large EMC values (Transmutation table & Condensers)

JamesTompkins:
Originally the EMC value label on the transmutation table would overflow into and past the forget slot as shown:

screen shot 2018-12-02 at 12 04 08 

However with modern modpacks such a stoneblock that includemods such as roost and RFDimensions, it is getting easier to surpass 1 billion EMC. To fix this I have added abbreviations once the EMC wealth surpasses 1 Billion to clean up the UI:

screen shot 2018-12-02 at 13 38 38

My code is easily expandable as there are two parallel arrays describing the abbreviations within TransmutationEMCFormatter.java. The number of decimal places are decided through the DecimalFormatter within constants.

### DIFF
--- a/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
+++ b/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
@@ -5,7 +5,7 @@ public class TransmutationEMCFormatter {
 		String postFix = "";
 		double magnitude = 1;
 
-		String formats[] = {"Q", "T","B"};
+		String formats[] = {"Quadrillion", "Trillion", "Billion"};
 		Double magnitudes[] = {1e15, 1e12, 1e9};
 
 		for (int i=0; i <formats.length; i++) {
@@ -17,6 +17,6 @@ public class TransmutationEMCFormatter {
 			}
 		}
 
-		return Constants.EMC_FORMATTER.format(EMC / magnitude) + postFix;
+		return Constants.EMC_FORMATTER.format(EMC / magnitude) + " " + postFix;
 	}
 }
--- a/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
+++ b/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
@@ -7,9 +7,9 @@ public class TransmutationEMCFormatter {
 		String postFix = "";
 		double magnitude = 1;
 
-		Double magnitudes[] = {1e18, 1e15, 1e12};
+		Double magnitudes[] = {1e12, 1e15, 1e18};
 
-		for (int i=0; i < magnitudes.length; i++) {
+		for (int i=magnitudes.length - 1; i >= 0; i--) {
 			double testMag = magnitudes[i];
 			if (EMC >= testMag) {
 				magnitude = testMag;
--- a/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
+++ b/src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
@@ -1,18 +1,19 @@
 package moze_intel.projecte.utils;
 
+import net.minecraft.client.resources.I18n;
+
 public class TransmutationEMCFormatter {
 	public static String EMCFormat(double EMC) {
 		String postFix = "";
 		double magnitude = 1;
 
-		String formats[] = {"Quintillion", "Quadrillion", "Trillion"};
 		Double magnitudes[] = {1e18, 1e15, 1e12};
 
-		for (int i=0; i <formats.length; i++) {
+		for (int i=0; i < magnitudes.length; i++) {
 			double testMag = magnitudes[i];
 			if (EMC >= testMag) {
 				magnitude = testMag;
-				postFix = formats[i];
+				postFix =  I18n.format("pe.emc.postfix." + i);
 				break;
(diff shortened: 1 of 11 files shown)


src/main/java/moze_intel/projecte/utils/TransmutationEMCFormatter.java
```diff
@@ -0,0 +1,22 @@
...
+	public static String EMCFormat(double EMC) {
+		String postFix = "";
+		double magnitude = 1;
+
+		String formats[] = {"Q", "T","B"};
+		Double magnitudes[] = {1e15, 1e12, 1e9};
+
+		for (int i=0; i <formats.length; i++) {
+			double testMag = magnitudes[i];
+			if (EMC >= testMag) {
```
MaPePeR: I'd do `testMag * 1000` here and also add M to the mix, so it always shows 3 digits. Maybe even `* 1000000`.

```suggestion
			if (EMC >= 1000  * testMag) {
```

pupnewfster:
Would potentially be useful to have this for the condenser as well (even though that only overflows the box after 10^17. One thing I think would be nice though is that if hovering over the abbreviated number it showed the exact amount. Given at times it is useful to see how much EMC you actually have.

JamesTompkins:
> Would potentially be useful to have this for the condenser as well (even though that only overflows the box after 10^17. One thing I think would be nice though is that if hovering over the abbreviated number it showed the exact amount. Given at times it is useful to see how much EMC you actually have.

I would add that myself if I was more familiar with the forge API

pupnewfster:
You should be able to do something along the lines of
```java
@Override
protected void renderHoveredTooltip(int mouseX, int mouseY) {
    if (mouseX > emcLeft && mouseX < emcRight && mouseY > emcTop && mouseY < emcBottom) {
        drawHoveringText(Arrays.asList(actualEMCasString), mouseX, mouseY);
    } else {
        super.renderHoveredTooltip(mouseX, mouseY);
    }
}
```
Though of course you would need to calculate the bounds of the text and pass the correct value for what the EMC string is, though you probably can just estimate it based on what the default font size is (ignoring custom fonts) and also how many digits/characters you are displaying.

sinkillerj:
I'm not opposed to the idea, but we should def implement said tooltip. It would also be nice if we could make this look a bit more elegant. Maybe "1.8 Trillion" instead of "T"? Open to suggestions here.

JamesTompkins:
To summarize the commits:
The tool tip has been added, abbreviations removed from suffixes, DecimalFormatter returned to normal. As a workaround for space the EMC label was separated and moved above the EMC value.

<1 Trillion:
screen shot 2019-01-29 at 21 56 39

\>= 1 Trillion:
screen shot 2019-01-29 at 21 56 51

Tooltip:
screen shot 2019-01-29 at 21 56 57

JamesTompkins:
(Apologies for all the commits)
Added abbreviations to condensers for higher values to prevent them overflowing too, acts exactly like transmutation table.

williewillus:
sorry for taking so long to get back to this

the changes look good for english, but
1. postfixes aren't translated/localizable to other languages
2. other languages (like east asian ones) don't use increments of 1000 for postfixes, they use 10000

given that, is there maybe a library we could find that could do this for us?

pupnewfster:
After doing a quick look, I was able to see a library that does this. I couldn't even find that some countries use 10000 instead of 1000 for postfix increments. (The closest thing on that subject I could find was that in 2003 the `International Bureau of Weights and Measures` declared the proper way to do it was in groups of three numbers separated by a thin space (not a comma or dot)).

Though I do agree that it makes sense to localize the postfix. I think the best way to do it would have it get the string by an index (right indexed based on the current code): So it would be `0` for Trillion, `1` for Quadrillion etc. Then add to the lang file
```
pe.number_postfix.0=Trillion
pe.number_postfix.1=Quadrillion
pe.number_postfix.2=Quintillion
```
And make the method return `return Constants.SINGLE_DP_EMC_FORMATTER.format(EMC / magnitude) + " " + I18n.format("pe.number_postfix." + index);` (The index would have to be calculated of course)

JamesTompkins:
I added support for translation and added to a few of the languages that seemed to translate well in google translate. 

I added it after 'TOO MUCH' in the translation files as it seemed most appropriate place for it. Ukrainian seemed to have words for everything except Quintilion so I left it English to be translated later.

pupnewfster:
I recommend reversing how you index the post fixes so that it goes from smallest to largest. (`0` being Trillion, etc). Just so that if we ever want to use the code on even larger numbers it is easy to update.

The simplest way may be to reverse your loop direction going down and then have the magnitudes go from smallest to largest. So something like:
```java
double[] magnitudes = {1E12, 1E15, 1E18};
for (int i = magnitudes.length - 1; i >= 0; i--) {
```
or
```java
double[] magnitudes = {1E12, 1E15, 1E18};
for (int i = magnitudes.length; i > 0; --i) {
```

sinkillerj:
There is no need to localize the TOO MUCH as its rendered unreadable anyway.

JamesTompkins:
>There is no need to localize the TOO MUCH as its rendered unreadable anyway.

@sinkillerj It was already there I didn’t add it, I just thought it would be a good place to add it as they’re somewhat related

williewillus:
looks good enough

PR #1793, opened 2018-12-17
Upgraded Forge to latest stable and mappings to final 1.12 mappings

pupnewfster:
A decision should be made about if how I changed build.gradle is fine or if it should be changed another way. I did it this way as it seemed it would ensure the most compatibility rather than changing `${minecraft_version}` from 1.12 to 1.12.2. When I attempted to change it, it would not build due to being unable to set things in update.json

I tested to make sure this still works on Minecraft 1.12 (and it does), so keeping the version number as 1.12 I think will be the least confusing.


pupnewfster:
You are right just adding that sufficed.

PR #1795, opened 2018-12-21
Tab Completions for commands.

DoctorFTB:
Added Tab Completions for 2 commands.
IDK what Tab Completions need for other commands.

### DIFF
--- a/src/main/java/moze_intel/projecte/network/commands/ClearKnowledgeCMD.java
+++ b/src/main/java/moze_intel/projecte/network/commands/ClearKnowledgeCMD.java
@@ -10,14 +10,23 @@ import net.minecraft.command.WrongUsageException;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.text.Style;
 import net.minecraft.util.text.TextComponentTranslation;
 import net.minecraft.util.text.TextFormatting;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 public class ClearKnowledgeCMD extends CommandBase
 {
+
+    public static final List<String> defaultSelectors = Arrays.asList("@p", "@a", "@r", "@e", "@s");
+
 	@Nonnull
 	@Override
 	public String getName()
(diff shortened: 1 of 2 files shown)


src/main/java/moze_intel/projecte/network/commands/ClearKnowledgeCMD.java
```diff
@@ -10,14 +10,23 @@
...
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 public class ClearKnowledgeCMD extends CommandBase
 {
+
+    public static final List<String> defaultSelectors = Arrays.asList("@p", "@a", "@r", "@e", "@s");
```
williewillus: make private and CONSTANT_CASE name

src/main/java/moze_intel/projecte/network/commands/ClearKnowledgeCMD.java
```diff
@@ -53,7 +62,20 @@ public void execute(@Nonnull MinecraftServer server, @Nonnull ICommandSender sen
 		}
 	}
 
-	@Override
+    @Override
+    public List<String> getTabCompletions(MinecraftServer server, ICommandSender sender, String[] args, @Nullable BlockPos targetPos) {
+        if (args.length == 1)
+        {
+            List<String> list = new ArrayList<>(Arrays.asList(server.getOnlinePlayerNames()));
```
pupnewfster: Personally I think if you are bothering to implement tab complete you should at least have it be Tab complete instead of tab overwrite like lots of mods implement because of being lazy. For your implementation say there are two people online, bob and alice. If I were to start typing "ali" and hit tab it would show `bob, alice, @p, @a, @r, @e, @s` instead of completing to the one finishing option of "alice". The same goes for the other command's tab complete. In addition if there are no arguments that is when you should be displaying this full list.

This is what I use in my mods for tab complete of player names:
```java
if (args.length == 0) {
    return Arrays.asList(server.getPlayerList().getOnlinePlayerNames());
}
if (args.length == 1) {
    String partialName = args[0];
    return Arrays.stream(server.getPlayerList().getOnlinePlayerNames()).filter(name -> name.startsWith(partialName)).collect(Collectors.toList());
}
```
I would recommend doing something similar (given this does not have the "defaultSelectors" in it).

pupnewfster: Personally I think if you are bothering to implement tab complete you should at least have it be Tab complete instead of tab overwrite like lots of mods implement because of being lazy. For your implementation say there are two people online, bob and alice. If I were to start typing "ali" and hit tab it would show `bob, alice, @p, @a, @r, @e, @s` instead of completing to the one finishing option of "alice". The same goes for the other command's tab complete. In addition if there are no arguments that is when you should be displaying this full list.

This is what I use in my mods for tab complete of player names:
```java
if (args.length == 0) {
    return Arrays.asList(server.getPlayerList().getOnlinePlayerNames());
}
if (args.length == 1) {
    String partialName = args[0];
    return Arrays.stream(server.getPlayerList().getOnlinePlayerNames()).filter(name -> name.startsWith(partialName)).collect(Collectors.toList());
}
```
I would recommend doing something similar (given this does not have the "defaultSelectors" in it).

PR #1805, opened 2019-01-18
Fix freeze when long overflows

pupnewfster:
Apparently I didn't test (or didn't test all cases) when a long can overflow and this fixes it so that Minecraft does not freeze if it does. Also removes comment that is from before the int -> long conversion which is not true for BigFraction

### DIFF
--- a/src/main/java/moze_intel/projecte/emc/arithmetics/HiddenBigFractionArithmetic.java
+++ b/src/main/java/moze_intel/projecte/emc/arithmetics/HiddenBigFractionArithmetic.java
@@ -7,22 +7,12 @@ public class HiddenBigFractionArithmetic extends FullBigFractionArithmetic
     @Override
     public BigFraction div(BigFraction a, long b)
     {
-        try
+        BigFraction result = super.div(a, b);
+        if (BigFraction.ZERO.compareTo(result) <= 0 && result.compareTo(BigFraction.ONE) < 0)
         {
-            if (this.isFree(a)) return getFree();
-            if (a.getNumerator().compareTo(MAX_LONG) > 0 || a.getDenominator().compareTo(MAX_LONG) > 0) {
-                //Overflowed a long as BigFraction can go past Long.MAX_VALUE
-                return BigFraction.ZERO;
-            }
-            BigFraction result = a.divide(b);
-            if (BigFraction.ZERO.compareTo(result) <= 0 && result.compareTo(BigFraction.ONE) < 0)
-            {
-                return result;
-            }
-            return new BigFraction(result.longValue());
-        } catch (ArithmeticException e) {
-            return BigFraction.ZERO;
+            return result;
         }
+        return new BigFraction(result.longValue());
     }
 
 }
--- a/src/main/java/moze_intel/projecte/emc/arithmetics/HiddenBigFractionArithmetic.java
+++ b/src/main/java/moze_intel/projecte/emc/arithmetics/HiddenBigFractionArithmetic.java
@@ -10,7 +10,7 @@ public class HiddenBigFractionArithmetic extends FullBigFractionArithmetic
         try
         {
             if (this.isFree(a)) return getFree();
-            if (a.getDenominatorAsLong() <= 0) {
+            if (a.getNumerator().compareTo(MAX_LONG) > 0 || a.getDenominator().compareTo(MAX_LONG) > 0) {
                 //Overflowed a long as BigFraction can go past Long.MAX_VALUE
                 return BigFraction.ZERO;
             }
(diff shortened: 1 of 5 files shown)


williewillus:
how does this freeze? it should just crash or bail

pupnewfster:
Before this change, when my pack was calculating things greater than long (or in a minipack for testing setting coal to max long) Minecraft would just freeze on attempting to load a world. No errors showing in the log. I let it sit for 5-10 minutes one time while I was still trying to figure out why my minecraft was freezing, and ended up giving up and force closing it.

pupnewfster:
This change also makes it follow a closer path to how it worked (the ArithmeticException would be thrown) while it was still an integer as per the comment I removed. Because there is also the chance that with how the code is currently if it overflows twice it could end up having the long value be positive again.

williewillus:
I'll let @MaPePeR look at this one

pupnewfster:
For reference from discord:

```the change it checking if the numerator or denumerator is greater than max long (the cases when arithmetic exception would have been thrown in the past was if either exceeded max int). What it currently does is try to check if the denominator "overflowed" and became negative. The issues with it is it could have overflowed so much that it is positive again, AND the bigger issue is that it doesn't check to see if the numerator is what overflowed.

BigFraction stores the numerator and denominator in BigInteger, so I am comparing two big ints to see if it overflowed (instead of attempting to "guess" by putting it into a long and seeing if I think it overflowed)```

**Comment by MaPePeR** [comment]
Would be great if you could write a test case for the case you are trying to fix.

With this change we are limiting the possible emc range, I think. An item that crafts 2 and would have an emc of (max long /2 + 1) couldnt get an emc value calculated, I think.

But it's 7 in the morning, so....

**Comment by pupnewfster** [comment]
I will write a test case in the morning (heading to bed given it is 1 am for me). This is similar functionality to the limitations we had back on integers, THOUGH given how BigFraction uses BigInteger internally I will look tomorrow to see if this works if I don't check denominator and only check the numerator. If I remember correctly the reason why I added the current check in the first place was that it was having some issue not meeting a test due to some error trying to find the GCD

**Comment by MaPePeR** [comment]
I also kind of doubt that the game just freezes. I assume there is an infinite loop somewhere. I think the code does not really do proper loop detection, but only fixes them by iterating a few times over them decreasing emc as it goes until 0 is reached. With biginteger this "few times" might actually be quite large.

**Comment by pupnewfster** [comment]
Ok @MaPePeR I added a test. I also tried reverting my fix, and the test failed. It appears that there already is a test to see what happens when an item crafts multiple things and I believe that works properly.

**Comment by pupnewfster** [comment]
I agree that in an ideal world it would be nice to have some sort of check that allows for the autogenerated EMC values to go above and then come back below max long. Though the current state of it not loading under certain conditions is even worse.

From the comment I am removing due to it not being the case for when that exception is thrown anymore. `the numerator or denumerator exceeds Integer.MAX_VALUE.` it would return zero. This pull makes the same be true, except with `Long.MAX_VALUE`. 

I tried removing the numerator check and only leaving the new denominator check, however the test I added fails without it.

I recommend that this be merged as it fixes the current issue, and that potentially a new issue is opened for requesting support for calculating EMC values of things that crafting recipe goes past the max long value. (Similar to how the hidden fraction stuff works for with glass panes). I do agree it would be a useful thing to have (and I can try to think of some way of doing it when I have a bit more time) but I believe it is somewhat out of the scope of this pull.

**Comment by pupnewfster** [comment]
Actually that being said, I just thought of a way that MAY work so I am doing some testing and if it does I will update this pull since I think I may be able to directly check the division result. After my initial testing now I think the current issue is in converting something that doesn't fit in a long to a long. Not actually that it is having issues figuring out the fraction.

**Comment by pupnewfster** [comment]
@MaPePeR if you could review again and merge/comment that it can be merged if acceptable that would be great. I modified the code (doesn't cause my minecraft the hang) and works with a depth size of one for going past overflow and returning to below it. At some point maybe we will want a HiddenOverflowArithmetic type thing that supports more depths.

The new code will work for the example that you asked about (I added two test cases to verify it does). However, what won't work is if I craft something that creates an item over EMC value of MAX_LONG and then have a recipe that uses two of those too large items to create 100 items that are below MAX_LONG, those 100 items will not have an EMC value.

File: src/main/java/moze_intel/projecte/emc/arithmetics/HiddenBigFractionArithmetic.java [hunk 1]
```diff
@@ -7,22 +7,12 @@
     @Override
     public BigFraction div(BigFraction a, long b)
     {
-        try
+        BigFraction result = super.div(a, b);
+        if (BigFraction.ZERO.compareTo(result) <= 0 && result.compareTo(BigFraction.ONE) < 0)
         {
-            if (this.isFree(a)) return getFree();
```
MaPePeR: I'd be surprised if we could skip this line. Otherwise it looks like this could work.

pupnewfster: We can skip it because it is covered by `super.div(a, b);` Because HiddenBigFractionArithmetic extends FullBigFractionArithmetic. So the isFree check already gets done there.

Issue #1777, opened 2018-11-30
[Suggestion] Add an option to disable Philostone entity randomizer projectile 


Saereth:
I'm Working on developing a pack for FTB that uses projecte tied in with progression but the philostone currently enables access to mobs from other dimension such as ghasts and blazes far before the player could ever get to those mobs. This becomes a pretty big issue for gating certain progression steps as it also bypasses gamestages/mobstaging. My only two options currently are disable the philostone entirely (Which we dont want to do as we're utilizing the inworld transmutations extensively) or use a hacky ASM override approach like this mod does:
https://minecraft.curseforge.com/projects/ceds-unleashed-control

I figured requesting an official config option would be best if you're open to it to avoid that sorta hackery. Thanks!

sinkillerj:
Ill look into this for the next version, I can certainly agree that this is a much needed balance tweak in some packs.

pupnewfster:
I think at some point when I get the chance I will look into adding support to the CraftTweaker integration to modify the peaceful and mob lists of the randomizer projectile. That way modpack makers will also be able to add mobs from other mods to be able to be randomized, in addition to removing/clearing the existing lists.

For quicker lookup when I get around to implementing this, the two lists are in WorldHelper.java and are called `peacefuls` and `mobs`

Saereth:
oh now thats very exciting

pupnewfster:
https://crafttweaker.readthedocs.io/en/latest/#Mods/ProjectE/EntityRandomizer/

PR #1812, opened 2019-01-29
Add CrT support for Philo Stone Entity Randomizer

pupnewfster:
Added a safety check to avoid an infinite loop in CollectionHelper if the list is of size 1 containing the element to exclude. In that specific case it ignores the exclusion request,

I will also add documentation later to the CrT docs documenting the different options and how to use the methods.


pupnewfster:
Forgot to link it via comment but this closes #1777

Issue #1773, opened 2018-11-27
[SUGGESTION] Add Watch of Flowing Time blacklists for randomization ticks and entity slowdowns.


EliteMasterEric:
The Watch of Flowing Time currently affects all blocks which utilize randomization ticks, and all living entities, within its radius, with no blacklists for modders or modpackers to add custom entities or blocks to.

The Watch should have blacklist API calls like the existing one, which modders can use, one which takes classes extending Block and one which takes classes extending EntityLiving. Classes in the former will be ignored when calling randomization ticks and classes in the latter will be called when slowing an entity's motion using the Watch.

Preferably this could also incorporate the suggestion in #1772. This would ultimately allow both modpackers and mod developers to exclude the Watch of Flowing Time from affecting certain items, to improve either stability or balance.

EliteMasterEric:
Looks like this got fixed a few years back

PR #1813, opened 2019-01-29
Improve performance for DM Pedestals and the Watch of Flowing Time.

iTitus:
Additionally fix a little bugs with the blacklists. They actually work now.


iTitus:
I can also try to tackly #1773 if you want

williewillus:
get rid of your whitespace changes please

iTitus:
> get rid of your whitespace changes please

Sorry about that, I thought everything is supposed to be formatted with tabs. Will change

iTitus:
It should be possible to look at the diff now ;)

pupnewfster:
You are still missing some whitespace difs
https://github.com/sinkillerj/ProjectE/pull/1813/files Has  +142, -115
https://github.com/sinkillerj/ProjectE/pull/1813/files?w=1 (Whitespace disabled) Has +125, -98

Also what is the reason for adding static to a lot of the things in TimeWatch.java

iTitus:
The rest of the whitespaces in the diff are changed indentation due to changed control flow.
I static-ed the methods because the simply don't need the item instance - so basically for style.

iTitus:
As you did a release recently @sinkillerj, you should know, that the blaclist for the watch of flowing time doesn't work, because the 'contains' logic is faulty.

williewillus:
why is the contains logic faulty?

the String[] array is an array of registry names, and we check Arrays.asList(arr).contains(<te's registry name>.toString()). Seems fine to me.

As for the original PR, I'd like to see some profiling to see if this actually gets better

iTitus:
> why is the contains logic faulty?

Strings were inserted into the blacklist using `Class.getName()`. But they were compared with `Class.toString()`. And `toString()` prepends `"class "` to the name, thus for any `Class<?> clazz` `clazz.getName() != clazz.toString()`.
See [static initializer](https://github.com/sinkillerj/ProjectE/blob/68bcbc80d24b6ba9be70f2883ce8ac1962b9050f/src/main/java/moze_intel/projecte/gameObjs/items/TimeWatch.java#L54-L57), [insertion](https://github.com/sinkillerj/ProjectE/blob/68bcbc80d24b6ba9be70f2883ce8ac1962b9050f/src/main/java/moze_intel/projecte/gameObjs/items/TimeWatch.java#L383), [comparison](https://github.com/sinkillerj/ProjectE/blob/68bcbc80d24b6ba9be70f2883ce8ac1962b9050f/src/main/java/moze_intel/projecte/gameObjs/items/TimeWatch.java#L209) and [`Class.toString()` reference](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#toString--).

iTitus:
And it was possible for users to remove the dm pedestal from the blacklist via the config and causing crashes with it. So I moved it to the hard-coded blacklist.

iTitus:
I profiled this setup:

Before: 1.3 - 1.5 µs
After: 1.0 - 1.3 µs
Which is a 0 - 50 % increase in tick speed ;)

PR #1818, opened 2019-02-26
Fully convert EMC over to Longs

pupnewfster:
Updated and continued version of #1724 

Some noteable changes since then:

- Found some more missing double to long changes, and reverted some changes to longs that were supposed to be doubles (the numbers were always `<= 1`
- Reverted/removed all formatting changes so that the dif is a lot easier to read.
- All balance changes removed, EXCEPT the RM furnace still requires 2 EMC instead of 1.6, AND the EMC per tick of the time watch has been increased by `0.5`.
- Create deprecated method variants of as many replaced API methods as possible.
  - The API changes that could not be deprecated
    - `double IKnowledgeProvider#getEmc()`
    - `double IEmcStorage#getStoredEmc()`
    - `double IEmcStorage#getMaximumEmc()`
    - `double IItemEmc#getStoredEmc(IItemStack)`
    - `double IItemEmc#getMaximumEmc(IItemStack)`
- Bumped API version number.

I believe this Pull Request is complete, but more testing can never hurt.

Edit: I cleaned up the description because thanks to #1827 and other changes in the pull a good part of it has been invalidated. GitHub keeps a dif of comments/descriptions so if for any reason it needs to be viewed it can be viewed with that.


williewillus:
no one outside of projectE should be implementing IEmcProxy

pupnewfster:
I know, but even so it is cleaner imo to have such a simple wrapper just be default instead of implementing the wrapper.

src/main/java/moze_intel/projecte/gameObjs/items/rings/SWRG.java
```diff
@@ -117,7 +119,11 @@ else if (stack.getTagCompound().getInteger(TAG_MODE) == 3)
 			toRemove = 0.64F;
 		}
 
-		removeEmc(stack, toRemove);
+		unprocessedEMC += toRemove;
+		if (unprocessedEMC > 1) {
+			removeEmc(stack, 1);
+			unprocessedEMC--;
+		}
```
williewillus: mutable state in an item is a huge no.

williewillus: same for every other item you do this

pupnewfster: It isn't static so it should be per item? Or is it because it is in Item, and that is on a lower tier than ItemStack, so they are all really using the same one.

williewillus: that's not how it works. every different itemstack of a swrg will share the same SWRG `Item` instance

williewillus: you probably need to put this in NBT

pupnewfster: Sounds good, I will take a look at it later today. I had forgotten that there is only one `Item` instance and that it wasn't like ItemStack where there are plenty of them.

pupnewfster: Fixed this, I also went ahead and made it so that it also stores the relay bonus, and partial EMC in collectors as well in NBT. Given initially I removed it for consistency and a number `< 1` was not really a big deal. I crossed out the previous text in the pull description.

PR #1822, opened 2019-02-28
Improvements to World Transmutation JEI integration

pupnewfster:
Moves the JEI recipe wrapper for world transmutation out of the WorldTransmutations class so as to not cause crashes without JEI installed. This could have been done with the Optional interface, except I decided to also revamp the recipe category so that it is more programatic for generating the list of items, rather than having lots of hard coded conditions. It also no longer shows a bunch of blank entries for transmutaions of alternate states of blocks (such as rotated logs or pumpkins).

### DIFF
--- a/src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
+++ b/src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
@@ -1,0 +1,153 @@
+package moze_intel.projecte.integration.jei.world_transmute;
+
+import mezz.jei.api.ingredients.IIngredients;
+import mezz.jei.api.recipe.IRecipeWrapper;
+import moze_intel.projecte.utils.WorldTransmutations;
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockStaticLiquid;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.init.Blocks;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fluids.FluidRegistry;
+import net.minecraftforge.fluids.FluidStack;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class WorldTransmuteEntry implements IRecipeWrapper
+{
+	private ItemStack inputItem = ItemStack.EMPTY;
+	private ItemStack leftOutputItem = ItemStack.EMPTY;
+	private ItemStack rightOutputItem = ItemStack.EMPTY;
+	private FluidStack inputFluid;
+	private FluidStack leftOutputFluid;
+	private FluidStack rightOutputFluid;
+	private boolean isValid = true;
+
+	public WorldTransmuteEntry(WorldTransmutations.Entry transmutationEntry)
+	{
+		Block inputBlock = transmutationEntry.input.getBlock();
+		IBlockState leftOutput = transmutationEntry.outputs.getLeft();
+		IBlockState rightOutput = transmutationEntry.outputs.getRight();
+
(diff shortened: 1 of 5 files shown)


src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
```diff
@@ -0,0 +1,158 @@
...
+			if (rightOutputFluid == null)
+			{
+				rightOutputItem = itemFromBlock(rightOutput.getBlock(), rightOutput);
+			}
+		}
+	}
+
+	private FluidStack fluidFromBlock(Block block)
+	{
+		if (block instanceof BlockStaticLiquid)
```
williewillus: what's the point if we just do == below anyway?

src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	@Override
+	@Nonnull
+	public java.util.List<String> getTooltipStrings(int mouseX, int mouseY)
+	{
+		if (mouseX > 67 && mouseX < 107 && mouseY > 18 && mouseY < 38)
+		{
+			return Collections.singletonList("Click in world, shift click for second output");
```
williewillus: translate

pupnewfster: Will do, that was part of the change I wanted to make when I editted the description this morning saying not to merge. Just by the time I got home I forgot about that change.

src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
pupnewfster: There isn't I will remove it. It was from the current handling except I have the other case elsewhere so it makes the added if statement not needed at all.

src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
```diff
@@ -0,0 +1,155 @@
...
+            return new FluidStack(FluidRegistry.WATER, 1000);
+        }
+        else if (block == Blocks.LAVA)
+        {
+            return new FluidStack(FluidRegistry.LAVA, 1000);
+        }
+		return null;
+	}
+
+	private ItemStack itemFromBlock(Block block, IBlockState state)
```
williewillus: why not just use statetostack?

src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
```diff
@@ -0,0 +1,155 @@
...
+        {
+            return new FluidStack(FluidRegistry.WATER, 1000);
+        }
+        else if (block == Blocks.LAVA)
+        {
+            return new FluidStack(FluidRegistry.LAVA, 1000);
+        }
+		return null;
+	}
+
```
williewillus: whitespace is a bit wack

src/main/java/moze_intel/projecte/integration/jei/world_transmute/WorldTransmuteEntry.java
pupnewfster: Because it does not work quite properly for this case. It gives "invalid" item stacks for things like rotated pumpkins, or non decayable leaves. Attempting to use pickBlock makes it get the proper type, though I did change the fallback to use statetostack, as I did not realize that was a method when I was writing this. (Discovered its existence while rewriting the mercurial eye)

Issue #650, opened 2015-04-14
[Suggestion] Extend mode for Mercurial Eye


Nova001Sol:
The Mercurial Eye, my alchemical must-have construction thing. I can build and transmute with it. The one thing I need, however, is a way to easily extend my walls and roof (oh, the ROOF!). Like ExUtils' Builder Wand, but Klein Star-powered. A Builder Wand is a fairly simple tool (as you probably know, but suggestions need detail) that allows one to quickly place blocks from their inventory. When right-clicked with on any side of any given block that the user has in their inventory, the Wand places blocks on every connected block of that same type on the side that was right-clicked on. It does only one layer at a time, and has a limited range. Connected blocks are adjacent blocks or diagonally connected blocks on the same plane that was originally right-clicked.

sinkillerj:
Maybe something to think about for the addons mod, as of right now however the Eye does have wall mode etc.

mFaunda:
Sinkillerj I almost explained to you how it used to work before I saw "owner" on your tag.  There's no extension mode in the version I just downloaded from CurseForge (ProjectE-1.12-PE1.3.0.jar - Apr 30, 2018).  Only normal mode and transmutation mode.  I want to make a floor bottoming off the END and I'm running into the same problem Nova is with their roof making.  Your "wall" (normal) mode still requires a block to be behind where you want the wall to actually go.  If you point me to the file path that tells the mercurial eye what it's doing I'd happily take a swing at it.

Heimdell:
I remember fixing extension mode myself and it being merged; I wonder, what happened after.

pupnewfster:
Closing this as it is included in 1.12 PE 1.4.1 (and in the 1.14 and 1.15 versions of ProjectE)

Issue #845, opened 2015-07-12
Fix/redo mercurial eye


williewillus:
I don't know if anyone even uses it as it is, but the Mercurial Eye is broken (technical-wise) in 1.8, which I have yet to fix. 

I looked at it and in addition to being broken, it doesn't even seem feature-complete with what it was in EE2. I tried using it ingame in 1.7 and it was super clunky and frequently placed blocks onto my head as well. So I'm going to redo it soon, properly. Probably for 1.8 because of recent code cleanups that would make it much easier than doing it in 1.7 (blockpos stuff)

HFSaphelion:
I have been trying to use it, but it will occasionally leave holes in the area I'm trying to fill. In my case specifically, I'm using it in SMP with a charged Klein Star Omega to fill in a huge @#$%ing hole that one of my server-mates dug into the world with a Destruction Catalyst. I'm trying to fill the hole in with regular stone blocks, but even fully powered up (where it should be making a 9x9 block square of stone), it will frequently leave empty holes in the wall which cannot be fixed with the mercurial eye. You HAVE to fill them by hand. I'm hoping this is the issue you are referring to and not another one. :/

Heimdell:
I used this thing a lot. I insist on properly implementing the extension mode.

Flexico:
I use this as well! I planned to build a castle with it, but without extension mode it's really clunky.

pupnewfster:
Closed as this got fixed when #1824 was merged

PR #1824, opened 2019-03-02
Mercurial Eye Rewrite

pupnewfster:
Rewrote the Mercurial Eye to have six modes. All four classic modes from EE2 are now included, as well as, more modern variants of the Extension and Transmutation modes. This pull is based off of the merceye branch and #1479, though lots of the logic has been rewritten/fixed/optimized. Also the feature from the original pull that added a feature on left click did not get added, as it felt clunky and unintuitive while testing.

Modes:
1. Creation: This mode is from EE2 and acts as if you are holding and placing the block stored in the key. It allows you to place blocks in replaceable blocks such as tall grass, and you can sneak to place against them instead. This mode has gained a new ability that it did not have in classic however, you can now right click in the air to place a new block there similar to how angel blocks work.

2. Extension: A modern rewrite/version of how the Extension mode worked in EE2. This mode works a lot more like builders wands do in that it brings the right clicked block's pattern towards the clicked side. It uses the item stored in the Mercurial Eye as a key, however if there is no "key" it will analyze the block clicked and attempt to use that to bring it forwards. Like the classic version charge levels allow it to change up to 1/9/25/49/81 blocks at a time.

3. Extension-Classic: The faithful extension mode from classic EE2. It behaves differently from how we currently think of extension building tools work. It fills in any gaps in up to a 9x9x1 box using the item stored in the Mercurial Eye. Like the modern variant, it has gained the ability to analyze the block clicked and use it instead if there is no stored item.

4. Transmutation: A modern rewrite/version of how the Transmutation mode worked in EE2. Instead of replacing all blocks in up to a 9x9x1 area of the clicked type, it replaces up to 81 touching blocks of the same type in the clicked plane. Like the classic version charge levels allow it to change up to 1/9/25/49/81 blocks at a time.

5. Transmutation-Classic: The faithful transmutation mode from classic EE2. It allows replacing blocks of a specific type in up to a 9x9x1 box, however these blocks do not have to be touching, they just have to have the same state.

6. Pillar: Works the same as it did in classic EE2. Creates a 3x3x3/6/9/12/15 pillar in the direction you clicked (opposite from the face of the block you clicked). This does not replace any blocks that were already there unless they are replaceable (like tall grass).

Small note-able things:
- Both transmutation modes drop the block if it does not have an EMC value.
- Both extension modes, and the pillar mode will not place blocks where an entity currently is.
- The Mercurial Eye now localizes the mode names instead of just having them be hard coded in.


pupnewfster:
This will close:
#650 
#845

PR #1825, opened 2019-03-02
Display mode on Philosopher Stone

pupnewfster:
Until I was looking at the formatting and to see if ItemMode supported localizing the mode names for #1824, I was not aware that the Philosopher's Stone even had modes. This makes it more clear by adding the current mode back into the tooltip. My guess is this was lost when it was overwritten to display the information about opening the crafting grid.


williewillus:
can you put it above the crafting grid tooltip?

PR #1826, opened 2019-03-02
Go past MAX_LONG for calculating stack EMC Tooltip

pupnewfster:
Shows numbers past MAX_LONG instead of obfuscated text for when the stack size is bigger than max long. This way people do not have to manually calculate how much EMC they have in a stack. Also given the tooltip is just a string it doesn't cause any other calculations to break by changing how the string is calculated.
image

### DIFF
--- a/src/main/java/moze_intel/projecte/utils/EMCHelper.java
+++ b/src/main/java/moze_intel/projecte/utils/EMCHelper.java
@@ -18,6 +18,7 @@ import net.minecraft.util.math.MathHelper;
 import net.minecraftforge.items.CapabilityItemHandler;
 import net.minecraftforge.items.IItemHandler;
 
+import java.math.BigInteger;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


src/main/java/moze_intel/projecte/utils/EMCHelper.java
```diff
@@ -307,6 +308,18 @@ public static String getEmcSellString(ItemStack stack, int stackSize)
...
+		if (EMCMapper.covalenceLoss == 1.0)
+		{
+			return " ";
+		}
+
+		BigInteger emc = BigInteger.valueOf(EMCHelper.getEmcSellValue(stack));
+
+		return " (" + Constants.EMC_FORMATTER.format(emc.multiply(BigInteger.valueOf(stackSize))) + ")";
+	}
+
```
williewillus: don't see the point of an extra method tbh, just make the one above use bigintegers too.

pupnewfster: Done

Issue #1750, opened 2018-09-02
Swiftwolf's Rending Gale - No Sound - 1.11.2


Craymozoid:
ProjectE Build: ProjectE-1.11.2-PE1.0.0B.jar

Forge Build: forge1.11.2-13.20.0.2253

Crash Log: n/a

Steps to reproduce:

1. Activate flight with "Swiftwolf's Rending Gale" in-game
2. Deactivate flight with "Swiftwolf's Rending Gale" in-game

What I expected to happen: peheal.ogg play over headset upon activation - peuncharge.ogg play over headset upon deactivation

What happened instead: No sound played.

pupnewfster:
Can be closed now

PR #1831, opened 2019-03-03
Adds back the sound for Swiftwolf's Rending Gale

pupnewfster:
This also adds it for the Ring of Arcana, when it is creating a SWRG projectile.

It mimics how Moving Minecarts have their sound done, so that the sound stops as soon as the projectile dies/hits something. It also gets quieter the further away it is.

I wasn't quite sure where to put the sound class, so if you have somewhere you would like it to be moved before this gets merged just let me know where to move it to.

Edit: ~Currently having the title include WIP while I try to fix a bug (that I remember also existing in EE2) where it plays the sound loudly and annoyingly when you are first opening up a world.~ Fixed


pupnewfster:
Closes #1750

Issue #1675, opened 2018-04-19
Dark/Red Matter Axe Not Working


Flexico:
Exact ProjectE version (do not say "latest", "latest on Curse", or similar): 1.12-PE1.2.0

Exact Forge version: 1.12.2-14.23.3.2659

Link to crash log (please use a paste site, do not attach the .txt or paste the log inline): https://pastebin.com/LAgktd2C

Steps to reproduce:

1. Hold Dark Matter or Red Matter Axe; right-click on tree

What I expected to happen: Large patch of forest chopped down

What happened instead: Nothing happens in-game, and the server spits out the above error (does not crash)

Normal69:
I was in the same situation, axe won't work, error looked like exactly yours.
Those tools need EMC energy in your inventory.
I've retried it with a stack of charcoal in my inventory, and:
IT WORKS ;)

PR #1833, opened 2019-03-03
Prevent crash if there is no tag for ItemMode

pupnewfster:
I was unable to reproduce the crash in #1675, but this should be a preventive measure to fix it crashing if somehow the tag compound becomes null.


williewillus:
getMode already accesses the compound, so this doesn't really do much

pupnewfster:
`getMode` has a check to see if it has a tag compound though. But doesn't create it if it is missing, so then changeMode still won't have one.

Issue #1747, opened 2018-08-22
[1.10.2] Explosives issue


Kaizodude:
**Tickets that do not conform to this template will be closed without comment**

Exact ProjectE version (do not say "latest", "latest on Curse", or similar): 1.10.2-PE1.3.1

Exact Forge version: Forge 12.18.3.2185

Link to crash log (please use a paste site, do not attach the .txt or paste the log inline): No crash log

Steps to reproduce:

1. Place A bunch of Nova Cataclysm

2. Light it

3. Let it explode 

What I expected to happen: Ores would pile up and Nova Cataclysm would be removed completely as it has been used

What happened instead: Ores DID pile up but every bit of Nova Cataclysm except one dropped back

PR #1834, opened 2019-03-03
Override correct method for block explosives.

pupnewfster:
Tested and this works now. Default implementation of `onBlockExploded` (not overridden by BlockTNT but is in Block)
```
world.setBlockToAir(pos);
onExplosionDestroy(world, pos, explosion);
```

The issue was that when the blocks got blown up, they were never setting it to air so just infinitely exploded. Now it overrides the same method as BlockTNT overrides to spawn the tnt entity.


pupnewfster:
This probably also fixes the issue from #1747

PR #1840, opened 2019-03-07
Handles changes to gamemode in regards to flight

pupnewfster:
This pull changes two things in regards to the flying checks.

- If the player was in a gamemode that allowed flying, and now they are not but `shouldPlayerFly()` is still true, it syncs the fact that they are allowed to fly to the client. (This allows players to not have to pickup the ring and put it back down to be able to fly again)
- If a player was in a gamemode that allowed flight, and no longer is set their `isFlying` to the same state as it was before. (Sometimes this has a micro drop of about 1/16 of a block before it makes them flying again, but this is preferable to them just not staying in flight at all)

The one quirk this change does not fix is if you are flying in survival and decide to run `/gamemode survival` then it will take you out of flight and you will still have to pickup the ring and put it back down to be able to start flying again.

### DIFF
--- a/src/main/java/moze_intel/projecte/handlers/InternalAbilities.java
+++ b/src/main/java/moze_intel/projecte/handlers/InternalAbilities.java
@@ -93,7 +93,7 @@ public final class InternalAbilities
 			wasFlyingGamemode = false;
 			wasFlying = false;
 		}
-		else //if (shouldPlayerFly())
+		else
 		{
 			if (!hadFlightItem)
 			{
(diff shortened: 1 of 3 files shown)


src/main/java/moze_intel/projecte/handlers/InternalAbilities.java
```diff
@@ -76,23 +79,40 @@ public void tick()
...
+
+				hadFlightItem = false;
 			}
-			
-			hadFlightItem = false;
+			wasFlyingGamemode = false;
+			wasFlying = false;
 		}
-		else if(shouldPlayerFly() && !hadFlightItem)
+		else //if (shouldPlayerFly())
```
williewillus: any reason why it's commented and not removed?

pupnewfster: Mainly just so it is clearer what that branch is at a glance. I can remove it if you wish.

williewillus: yeah, remove it - it's visible by just looking at the if condition

pupnewfster: Done.

PR #1858, opened 2019-04-03
Added ability to register items with NBTs and allowed for storing custom EMC in separate files

JPANv2:
One of the biggest issues with this mod was the lack of NBT support for items. Items could be registered even when they deeply depend on an NBT, but the result would be an empty item of that kind. One such example of this problem is Minecraft's Potions and Enchantment Books. Learning one of those items in the Transmutation table would result in an empty Enchantment Book, or in a Uncraftable Potion. This problem goes even further with mod support, with Forge-generated buckets producing no liquids or Tinkerer's Construct's Tools and Armor reverting to blank, useless versions of themselves.

This pull provides a simple solution to these problems: either through an API or through the two new "setEMCNBT" and "removeEMCNBT", it is possible to allow users to set their own values to items that should have their NBT stored, and have multiple, different values for each and every variation they can get their hands on. 

This pull also adds automatic knowledge generation for encountered Enchantment Books, however the same cannot be said of potions, even though a solution could be coming if this pull is not outright rejected.

I am aware of the problem of NBTs used as "damage" values (IE: energy capacity), and a solution was pushed forward: allow the users to "Blank Out" the NBT fields they think are not needed, such as the current energy of batteries and the like.

PS: Another small problem that comes with using a varied set of mods while playing is that the emc_custom file gets very bloated, and there is no easy way to share values from specific mods to other players without trying to merge two very large files. For that, I added a config option that allows ProjectE to store the custom emc in files separated by mod, allowing for easy sharing of custom emcs and less bloat than reusing a custom_emc.json in multiple minecraft instances.

### DIFF
--- a/src/main/java/moze_intel/projecte/api/proxy/IItemNBTEmcCalculator.java
+++ b/src/main/java/moze_intel/projecte/api/proxy/IItemNBTEmcCalculator.java
@@ -1,0 +1,80 @@
+package moze_intel.projecte.api.proxy;
+
+import java.util.Collection;
+
+import org.apache.commons.math3.fraction.BigFraction;
+
+import moze_intel.projecte.emc.arithmetics.IValueArithmetic;
+import net.minecraft.item.ItemStack;
+
+/**Interface that allows for calculating custom EMC values for non-defined entries at request time.
+ * 
+ * This interface allows for both creating new EMC values for existing items based on their NBT tags,
+ * as well as provide modifiers to the value if certain NBT Tags are present (such as enchantments)
+ * */
+
+public interface IItemNBTEmcCalculator{
+	
+	public static enum Operation{
+		OP_ADD{
+			public BigFraction operate(BigFraction originalEMC, BigFraction modifier){
+				return originalEMC.add(modifier);
+			}
+		},
+		OP_SUBTRACT{
+			public BigFraction operate(BigFraction modifier, BigFraction originalEMC){
+				return originalEMC.subtract(modifier);
+			}
+		},
+		OP_MUlTIPLY{
+			public BigFraction operate(BigFraction modifier, BigFraction originalEMC){
+				return originalEMC.multiply(modifier);
+			}
+		},
+		OP_DIVIDE{
+			public BigFraction operate(BigFraction modifier, BigFraction originalEMC){
+				return originalEMC.divide(modifier);
+			}
+		},
+		OP_SET{
+			public BigFraction operate(BigFraction modifier, BigFraction originalEMC){
+				return modifier;
+			}
+		},
+		OP_NONE;
+		
+		public BigFraction operate(long originalEMC, BigFraction modifier){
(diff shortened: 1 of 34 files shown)


MaPePeR:
There already is a way to share mod specific emc files using the custom conversion files. Does your pull request also work with these?

JPANv2:
I haven't changed anything from the custom conversions, just added an option to output/read the custom_emc.json from different files, separated by mods. So, custom conversions still work as in the previous versions.

JPANv2:
With this commit, I added a way to calculate potion EMC from the minecraft's PotionHelper class, allowing most NBT items from vanilla minecraft to have associated values. Also gave a custom value of 512 to Dragon Breath, so that lingering potions and tipped arrows could also have EMC.

Besides that, I added two API files, the IItemNBTFilter and the IItemNBTEmcCalculator, both to allow modders (and projectE itself) to filter undesired NBT tags and dynamically calculate EMC for items, generally based on their tags. 

IItemNBTFilter is used to create copies of an ItemStack with the desired NBTs, and filter out undesired ones. In this commit I also add four classes implementing it: the ItemNBTNameTagCleaner removes names from items, the ItemNBTStoredEMCCleaner cleans the StoredEMC tag from itemStacks, ItemNBTProjectEActiveModeCleaner cleans all other non-identifying ProjectE-defined NBT tags (Active and Mode) and ItemNBTEnchantmentTagCleaner which is not part of the registered Filters, and removes enchantments from non-book items (to fulfill the learnEnchantedItems = false option of the config.)

IItemNBTEmcCalculator allows for dynamically altering EMC of items without altering their knowledge. This can be used to add items with different EMC values where adding to the emc Map would not be practical, such as with TConstruct's pickaxes. Currently the mod implements three instances of the EMC Calculator:  EnchantmentCalculator calculates EMC of enchantment-containing items  based on the old EMCHelper.getEnchantEmcBonus() function; DurabilityCalculator calculates an EMC multiplier based on how much durability a tool has and StoredEMCCalculator takes the "StoredEMC" tag contents and adds it to the item EMC.

In order to add new Filters and Calculators to be run on knowledge checks, they must be registered to the ItemStackNBTManager class, with their respective register commands.

williewillus:
can you retarget this to 1.13? It's really late in 1.12's lifecycle to get such a major change to the core component of PE (emc mapping), I'd rather keep that version nice and stable. But from preliminary glances it pretty good.

williewillus:
Also, can you add unit tests?

JPANv2:
Sure, I can try to port it to the 1.13 version and add unit tests.  I'll add a new pull request when it's done and tested.

JPANv2:
Will now close this pull request, as the 1.13 version is ready.

UnicorNora:
I know this is an old PR but i would appreciate seeing this released for 1.12.2 or if someone on the ProjectE team could give me permission to release a fork with this Fix

PR #1928, opened 2019-10-23
Convert Player EMC to being a BigInteger

pupnewfster:
There is still a bit more cleanup I should do if we want to end up having this be a thing, but I am opening this draft PR so that we can try to decide if we want to make the Player's EMC be a BigInteger.

### Notable Changes
- Player EMC is a BigInteger
- The code now acts like there is no "max" value for the player's EMC. This removes a bunch of different checks for making sure that the EMC can still fit.
- Some adding/removal of EMC logic was changed to multiply and put the entire stack at once instead of looping and attempting to fit it for each stack. (This is because of there being no "max")

I do not think it makes sense to change the EMC for items, or tiles to go past max long as that would just be an absurd amount of EMC that is hard to reach for a single item even purposely inflating EMC values. (Though it may make sense to look into making it be an unsigned long). In some mod packs however having very large numbers for items make cause it so that the player reaches the limit for how much they have stored, and this change basically removes that limit.

Note: I have not done any profiling as of yet on the performance of this compared to the old system.


williewillus:
not sure if this makes sense from a consistency point of view tbh
performance wise I doubt you'll find anything significant

PR #2041, opened 2020-07-18
tehnut.info/maven is down, get HWYLA and CraftTweaker from CurseForge

xnrand:
This fixes the gradle build on my machine.

reference: TeamPneumatic/pnc-repressurized@f2a93f4c882e9258839c1049829dc47da4616cd5

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,7 @@ buildscript {
 
 plugins {
     id "com.github.johnrengelman.shadow" version "1.2.4"
+    id "com.wynprice.cursemaven" version "2.1.5"
     id 'java'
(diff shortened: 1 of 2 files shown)


pupnewfster:
There is no reason to be switching things over to cursemaven (especially the old deprecated gradle plugin instead of https://www.cursemaven.com/ Also I am pretty sure that there is a new maven for hwyla with this path `https://maven.tehnut.info/` or something like it as the base which would be the more proper thing to use.

Chew:
Hello, the new maven repo for tehnut.info/maven is https://maven.tehnut.info, this fixed the HWYLA error for me.

PR #2059, opened 2020-09-29
Updated to forge to stable 34.1.0 version and other Dependencies to latest versions as well as Updated to latest MCP and Fixed all current Mapping issues 

raziel23x:


gradle.properties
```diff
@@ -4,18 +4,18 @@ org.gradle.jvmargs=-Xmx3G
...
 minecraft_version=1.16.3
-mappings_version=20200723-1.16.1
+mappings_version=20200916-1.16.2
 loader_version=34
-forge_version=34.0.1
+forge_version=34.1.0
 # This determines the minimum version of forge required to use ProjectE
 # Only bump it whenever we need access to a feature in forge that is not available in earlier versions
-min_forge_version=34.0.0
+min_forge_version=34.1.0
```
pupnewfster, line 12: There is no need to bump the min forge version unless we need something from it.

build.gradle
```diff
@@ -189,8 +189,8 @@ dependencies {
     testImplementation "org.junit.jupiter:junit-jupiter-api:${junit_version}"
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:${junit_version}"
 
-    compileOnly fg.deobf("mezz.jei:jei-1.16.2:${jei_version}:api")
-    runtimeOnly fg.deobf("mezz.jei:jei-1.16.2:${jei_version}")
+    compileOnly fg.deobf("mezz.jei:jei-1.16.3:${jei_version}:api")
+    runtimeOnly fg.deobf("mezz.jei:jei-1.16.3:${jei_version}")
```
pupnewfster, line 193: This should go back to `${minecraft_version}` instead of 1.16.3

src/main/java/moze_intel/projecte/gameObjs/gui/GUITransmutation.java
```diff
@@ -12,6 +12,7 @@
 import moze_intel.projecte.utils.TransmutationEMCFormatter;
 import moze_intel.projecte.utils.text.PELang;
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.screen.AbstractCommandBlockScreen;
```
pupnewfster, line 15: ??

raziel23x: was a mistake will remove it

PR #2064, opened 2020-11-01
Fix bug where dedicated server crashed when igniting Nova Cataclyst and Nova CataClysm

unizLike:


pupnewfster:
Manually went ahead and cherry picked this, because for some reason when I rebased to grab the changes sin made to the 1.15 branch github/my IDE had some sort of stroke. But thanks for this I did miss that the constructor is (or now is) client side only.

PR #2079, opened 2021-01-18
Improve handling of combined fake ingredients

pupnewfster:
I believe this addresses what you were mentioning on discord @MaPePeR regarding the duplicate conversions for identical NormalizedSimpleStacks, as well as the duplicated recipes in cases when the matching stacks may be in a different order. Opening this as a PR for you to review when you get a chance.

### DIFF
--- a/src/api/java/moze_intel/projecte/api/mapper/recipe/INSSFakeGroupManager.java
+++ b/src/api/java/moze_intel/projecte/api/mapper/recipe/INSSFakeGroupManager.java
@@ -1,0 +1,22 @@
+package moze_intel.projecte.api.mapper.recipe;
+
+import java.util.Set;
+import moze_intel.projecte.api.nss.NormalizedSimpleStack;
+import net.minecraft.util.Tuple;
+
+/**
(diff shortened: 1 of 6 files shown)


MaPePeR:
Looks good to me. At first I thought that it would be better to place the "add conversions for fakegroup" code to the `NSSFakeGroupManager `, but that would also mean moving the container handling into that code and different mappers might behave differently and not need that, so I guess how its done here s is the better solution. 👍 

This should greatly reduce the needed RAM, calculation speed of the emc mapping and the filesize of the conversion-dump.

pupnewfster:
Ya, I tried to make it generic enough with the NormalizedSimpleStacks given I realized that it may be of use for addons making custom mappers, as they may be using NSSFluids or other custom NormalizedSimpleStacks that then have to be combined. Not to mention one thing I am thinking of doing at some point would potentially "kill off" the container handling how it is done now anyways as strictly speaking recipes have a "getRemainingItems" method which means custom recipes may not necessarily have containers returned, or may have other items "treated" as containers and be returned regardless. Though maybe for performance (and to reduce copying) I will only actually do that if the class is not a vanilla one so has a custom implementation.

PR #2098, opened 2021-02-24
Adjust Bronze EMC value

SDPhantom:
Alloying results in an output quantity of 4, not 1.

### DIFF
--- a/src/datagen/generated/data/projecte/pe_custom_conversions/metals.json
+++ b/src/datagen/generated/data/projecte/pe_custom_conversions/metals.json
@@ -34,6 +34,7 @@
       {
         "propagateTags": true,
         "output": "#forge:ingots/bronze",
+        "count": 4,
         "ingredients": {
           "#forge:ingots/copper": 3,
           "#forge:ingots/tin": 1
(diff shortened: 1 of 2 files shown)


pupnewfster:
Should be adjusted here: https://github.com/sinkillerj/ProjectE/blob/mc1.16.x/src/datagen/java/moze_intel/projecte/common/PECustomConversionProvider.java#L41 and then have `runData` be ran again or it will be reverted the next time data generation is ran. (If you aren't able to do so I can probably make this change sometime later today)

SDPhantom:
I haven't dug around to see how your builder works unfortunately. I was looking at examples of tag references to build a datapack for another mod and came across the calculation error.

pupnewfster:
This is still not what I asked for as you did not actually run the runData task as the hash for if the file is up to date did not end up changing.

SDPhantom:
Not sure how to do that tbh.
I only spotted the calculation error from looking at how to reference tags for building a datapack for another mod.

pupnewfster:
That is fine, and I am glad you spotted it, though I probably will then just end up closing this PR a bit later and making the one line change myself.

pupnewfster:
Turns out I had forgotten to actually have the file I was complaining about commit anyways, but fixed that along with this in that commit.

Issue #2071, opened 2020-12-08
Fuel textures off by a pixel


triffman33:
Exact ProjectE version:  1.14+

Steps to reproduce:

1. Load Game
2. Look at the Alchemical Coal, Mobius Fuel and Aeternalis Fuel
3. Notice its one pixel to the right

What I expected to happen: Be in the centre so it fits over the dark/red matter furnace slot outline of the fuel

What happened instead: One pixel to the right of the outline
![Untitled](https://user-images.githubusercontent.com/31788797/101530376-b979c480-3989-11eb-93b1-d695e052fafa.png)

Issue #2110, opened 2021-03-30
Duplication bug


malthenon:
**Tickets that do not conform to this template will be closed without comment**

version: 1.16.5-PE1.0.1B

Forge version: 36.0.46

Link to crash log (please use a paste site, do not attach the .txt or paste the log inline): NA

Steps to reproduce:

1. get any tier of antimatter relay and any tier klein star
2. place the klein star into the charging port of the antimatter relay.
3. break relay
4. relay and *two* klein stars drop.

What I expected to happen:

expected only one star to drop from breaking the relay.

What happened instead:

recieved an extra klein star.

PR #2135, opened 2021-07-06
Fixed fuel textures being offset by a pixel

wowmom98:
To fix issue #2071
I'm assuming the files in src/main/resources/assets/projecte/textures/item/fuels are all I should need to change, if it isn't let me know.


wowmom98:
I also went ahead and updated the build.gradle because it was necessary for me to get gradle working. I'm not 100% sure what I'm doing however so if for some reason this doesn't make sense let me know.

wowmom98:
Fixed another issue, this time issue #2110 . This removes a redundant override in the Relay class, because it was running itself and then it's superclass's, dropping the item in its charge slot twice. The override wasn't doing anything different to the superclass anyway, so it was entirely redundant.
However while fixing this issue I discovered a similar one with collectors, and I wrote about it in issue #2138.

pupnewfster:
Could you please rebase this PR to not have the build.gradle changes. Strictly speaking the texture fixes and the relay dupe fix should be in two separate PRs but it is fine if you want to just rebase this PR and leave them in one.

PR #2161, opened 2021-11-11
Fix handheld item model transforms due to wrong parent model

KitsuneAlex:
Also exclude datagen cache and update ForgeGradle, because otherwise the build breaks completely.


pupnewfster:
Ended up just doing it myself after I fixed the build.gradle things as the approach to methods in this was a bit odd but more importantly the datagen cache should not be excluded and there were a bunch of other weird removals from the build.gradle

Issue #2133, opened 2021-07-01
Incorrect EMC prediction


Apceniy:
**Tickets that do not conform to this template will be closed without comment**

Exact ProjectE version (do not say "latest", "latest on Curse", or similar): 1.16.5-PE1.0.1B

Exact Forge version: 36.1.0

Link to crash log (please use a paste site, do not attach the .txt or paste the log inline): No crash

Steps to reproduce:

1. Remove EMC for the philosopher's stone

What I expected to happen: Conversion recipes such as iron to gold, coal to alchemical coal will still predict EMC because philosopher's stone isn't consumed when crafting

What happened instead: Result items no longer have EMC if not set in other ways

MaPePeR:
Easiest fix might be to filter the `IngredientMap` for 0 entries [here](https://github.com/sinkillerj/ProjectE/blob/10fdd41f26d1dca4183e1ebcb74c990bf2d17a96/src/main/java/moze_intel/projecte/emc/IngredientMap.java#L16).

pupnewfster:
I can't reproduce this by just removing the EMC of the philo stone, it seems to still have an EMC value regardless until I remove the EMC of all ways of creating it.

@MaPePeR It filters it [here](https://github.com/sinkillerj/ProjectE/blob/mc1.16.x/src/main/java/moze_intel/projecte/emc/SimpleGraphMapper.java#L183-L186), but not [here](https://github.com/sinkillerj/ProjectE/blob/mc1.16.x/src/main/java/moze_intel/projecte/emc/collector/MappingCollector.java#L50). Do you think I should still add the filter earlier as it may make it so it has to do less calculations as it gets further into the mapper or just leave it be? Also do you think the second linked spot should skip adding it as a "usage" for the ingredient if there is none? (Note: Those two aren't mutually exclusive as there are other places or custom mappers/jsons where a conversion may end up having a specified size of zero so we may as well handle it properly in the mapper as well like we do now)

MaPePeR:
Adding the filter to the [second spot](https://github.com/sinkillerj/ProjectE/blob/mc1.16.x/src/main/java/moze_intel/projecte/emc/collector/MappingCollector.java#L50) sounds like a good idea to me, as it may indeed speed up the mapper and do less calculations.

[...]

PR #2169, opened 2022-01-11
Slight cleanup to conversion validation and potentially performance

pupnewfster:
I addressed the first few things in https://github.com/sinkillerj/ProjectE/issues/2133#issuecomment-1008120735, but for the last part about the loop, if I understand correctly:
- for all new ingredients it is getting any existing uses for the ingredient
- removing the old conversion for them.

If it looped the old conversion then it would be removing conversions that might not exist with the new one that is overwriting it I think? So I left it as is for now as I think that may be intended.


pupnewfster:
CC @MaPePeR as it isn't letting me use the request review feature to notify you via github.

MaPePeR:
Looks good to me. 👍

PR #2189, opened 2022-04-20
Add New options for EMC Display inside Energy Condensers

The-Silver-Kid:
Adds a configurable option to allow Users to see how much "Extra" EMC they have stored in their Energy Condensers.

Mode 1 will just cause the number to turn green when over the required amount.

Mode 2 will add the Stored amount next to the required amount. If the length is too long it will instead just say overflow.

If the stored is less then the required it wont do anything different. (right side in the image).

If you don't want to include this thats fine. Just something I think may be a bit handy while playing with high EMC valued items.

EMC modes

### DIFF
--- a/src/main/java/moze_intel/projecte/config/ClientConfig.java
+++ b/src/main/java/moze_intel/projecte/config/ClientConfig.java
@@ -1,6 +1,7 @@
 package moze_intel.projecte.config;
 
 import moze_intel.projecte.config.value.CachedBooleanValue;
+import moze_intel.projecte.config.value.CachedIntValue;
 import net.minecraftforge.common.ForgeConfigSpec;
 import net.minecraftforge.fml.config.ModConfig;
 
@@ -17,6 +18,7 @@ public class ClientConfig extends BasePEConfig {
 	public final CachedBooleanValue statToolTips;
 	public final CachedBooleanValue pedestalToolTips;
 	public final CachedBooleanValue pulsatingOverlay;
+	public final CachedIntValue condenserEmcDisplayMode;
 
 	ClientConfig() {
 		ForgeConfigSpec.Builder builder = new ForgeConfigSpec.Builder();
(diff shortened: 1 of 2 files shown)


src/main/java/moze_intel/projecte/gameObjs/gui/AbstractCondenserScreen.java
```diff
@@ -39,9 +40,40 @@ protected void renderBg(@NotNull PoseStack matrix, float partialTicks, int x, in
...
+			emc = TransmutationEMCFormatter.formatEMC(toDisplay);
+			this.font.draw(matrix, emc, 140, 10, 0x404040);
+			
+			if (menu.displayEmc.get() > menu.requiredEmc.get()) {
+				int xOff = 140 + this.font.width(emc);
+				emc = TransmutationEMCFormatter.formatEMC(menu.displayEmc.get());
+				if ((xOff + this.font.width(" / " + emc.getString())) < this.getXSize()) {
+					this.font.draw(matrix, " / " + emc.getString(), xOff, 10, 0x008000);
+				} else {
+					this.font.draw(matrix, " / Overflow", xOff, 10, 0x008000);
```
pupnewfster, line 72: All of this should be done via translation components defined in `PELang` and `PELangProvider`.

src/main/java/moze_intel/projecte/config/ClientConfig.java
```diff
@@ -40,6 +42,9 @@ public class ClientConfig extends BasePEConfig {
 		pulsatingOverlay = CachedBooleanValue.wrap(this, builder
 				.comment("The Philosopher's Stone overlay softly pulsates")
 				.define("pulsatingOverlay", false));
+		condenserEmcDisplayMode = CachedIntValue.wrap(this, builder
+				.comment("Sets how EMC is displayed inside Energy Condensers.\n0: Shows how much EMC is stored or how much is required to make the item if there is Excess stored EMC. (EE2 Style)\n1: Will show the stored amount of EMC and render green if the value exceeds required amount.\n2: If the stored EMC value is greater then required it will display the total EMC stored seperately in green.\nNote: If the value displayed would overrun the edge of the GUI it will instead state 'Overflow'.")
+				.define("showTotalEMCInCondenser", 0));
```
pupnewfster, line 47: I feel like this would make more sense just using an enum for the config. General handling would be making a new config value type that basically is a copy of https://github.com/mekanism/Mekanism/blob/1.18.x/src/main/java/mekanism/common/config/value/CachedEnumValue.java. I also believe you can pass multiple strings to `comment` to properly apply them on new lines rather than injecting manual `\n` declarations

src/main/java/moze_intel/projecte/gameObjs/gui/AbstractCondenserScreen.java
The-Silver-Kid: Ah Right. I'll fix that.

src/main/java/moze_intel/projecte/config/ClientConfig.java
The-Silver-Kid: Fair Enough. Ill adjust that too along with the Translations.

src/main/java/moze_intel/projecte/gameObjs/gui/AbstractCondenserScreen.java
The-Silver-Kid: How should i handle the `" / "` additions? Should i add translation keys for that too?

pupnewfster: Yes that way for en_ud it can reverse the order of things properly.

Issue #2024, opened 2020-06-01
Suggestion: A way to query, set or change a player's EMC using commands


James103:
**Tickets that do not conform to this template will be closed without comment**

Exact ProjectE version (do not say "latest", "latest on Curse", or similar): ProjectE-1.15.2-PE1.0.3B
(Note: This also applies for 1.12.2 as well)

Exact Forge version: 31.2.0

Link to crash log (please use a paste site, do not attach the .txt or paste the log inline): No crash

Steps to reproduce:

1. `/projecte setPlayerEMC <player> <integer>`
2. `/projecte addPlayerEMC <player> <integer>`
3. `/projecte queryPlayerEMC <player> [divisor]`

What I expected to happen:
The first command sets the EMC value for the specified player to the given value, and the second command changes the EMC value for the specified player by the given value (can be positive or negative). In both cases, the player's EMC value is bounded to between 0 and either `Long.MAX_VALUE` or `Double.MAX_VALUE`. For example, assuming you start with 1,000 EMC, running the command `/projecte setPlayerEMC @p 1100` is the same as running `/projecte addPlayerEMC @p 100`, with both setting your EMC to 1,100. Similarly, if you start with 1,000 EMC, running the command `/projecte addPlayerEMC @p -100` will deduct 100 EMC from you, leaving you with 900 EMC.

The third command queries the EMC value for the specified player. For example, if you have 1,000 EMC and you run `/projecte queryPlayerEMC @p` on yourself, it will return 1,000. Note that if you have more than 2,147,473,647 EMC when you run the `/projecte queryPlayerEMC @p` command, it will just return 2147483647, unless you specify a power of 10 to divide the return value by using the `[divisor]` argument. For example `/projecte queryPlayerEMC @p 3` returns 1,234 if you have 1,234,567 EMC, while  `/projecte queryPlayerEMC @p 18` returns 602,000 if you have 602 sextillion EMC.

What happened instead:
All of the three above commands don't exist yet.

Example use cases:
* To give the player a penalty to their EMC when he/she dies or takes damage
* To allow for various features that use EMC to heal/feed yourself
* To allow for you to summon entities using your EMC without using spawn eggs
* To allow for various things to give you bonus EMC
* To allow players to start with a specified amount of EMC without having them start with a Transmutation Table/Tablet and/or transmutable items

sinkillerj:
[...]

PR #2245, opened 2022-10-10
Fully Featured EMC & Knowledge Subcommands

DonovanDMC:
`/projecte emc`
- `add`, adds emc - removes if less than zero
- `remove`, removes emc - adds if less than zero
- `set`, sets emc
- `get`, gets emc
- `test` - tests if an amount can be removed from the player's emc

`/projecte knowlege`
- `clear`, clears knowledge (slightly different from `/projecte clearknowlege`, it returns a fail response if they have no knowledge)
- `learn`, adds the specified item to their knowledge, if not already learned
- `unlearn`, removes the specified item fro their knowledge, if learned
- `test`, tests if the user has the specified item in their knowledge

This is based on the [`/emc`](https://github.com/DonovanDMC/ProjectExpansion/blob/1.19.2/src/main/java/cool/furry/mc/forge/projectexpansion/commands/CommandEMC.java) and [`/knowledge`](https://github.com/DonovanDMC/ProjectExpansion/blob/1.19.2/src/main/java/cool/furry/mc/forge/projectexpansion/commands/CommandKnowledge.java) commands from Project Expansion. I've slightly modified them to better fit ProjectE.
Primary differences:
- Not differentiating between executing actions on other users and the executor
- Not attempting to send notifications

I couldn't actually find the datagen task (I didn't really try that hard), so I'm unsure if I've actually added all that was needed (I've just added the strings to the en_us lang manually)


src/main/java/moze_intel/projecte/PECore.java
```diff
@@ -48,12 +48,7 @@
 import moze_intel.projecte.network.PacketHandler;
 import moze_intel.projecte.network.ThreadCheckUUID;
 import moze_intel.projecte.network.ThreadCheckUpdate;
-import moze_intel.projecte.network.commands.ClearKnowledgeCMD;
-import moze_intel.projecte.network.commands.DumpMissingEmc;
-import moze_intel.projecte.network.commands.RemoveEmcCMD;
-import moze_intel.projecte.network.commands.ResetEmcCMD;
-import moze_intel.projecte.network.commands.SetEmcCMD;
-import moze_intel.projecte.network.commands.ShowBagCMD;
+import moze_intel.projecte.network.commands.*;
```
pupnewfster: For the most part in this project we use fully qualified imports rather than wildcards.

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+    private enum ActionType {
+        ADD,
+        REMOVE,
+        SET,
+        GET,
+        TEST
+    }
+    public static ArgumentBuilder<CommandSourceStack, ?> register(CommandBuildContext context) {
+         return Commands.literal("emc")
+            .requires(PEPermissions.COMMAND)
```
pupnewfster: This potentially should be its own node. At the very least the base command node is already covered by the parent command.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+public class KnowledgeCMD {
+    private enum ActionType {
+        LEARN,
+        UNLEARN,
+        CLEAR,
+        TEST
+    }
+    public static ArgumentBuilder<CommandSourceStack, ?> register(CommandBuildContext context) {
+        return Commands.literal("knowledge")
+            .requires((source) -> source.hasPermission(2))
```
pupnewfster: This potentially should be using permission nodes.

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+            .then(Commands.literal("add")
+                .requires(PEPermissions.COMMAND_EMC_ADD)
+                .then(Commands.argument("player", EntityArgument.player())
+                    .then(Commands.argument("value", StringArgumentType.string())
+                        .executes((ctx) -> handle(ctx, ActionType.ADD))
+                    )
+                )
```
pupnewfster: This and the others below it could likely be a `.then(executeWithParameters(ActionType))` with a helper method `executeWithParameters(ActionType)` method that does:
```java
return Commands.argument("player", EntityArgument.player())
    .then(Commands.argument("value", StringArgumentType.string())
        .executes(ctx -> handle(ctx, actionType))
   );
```

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static int handle(CommandContext<CommandSourceStack> ctx, ActionType action) throws CommandSyntaxException {
+        ServerPlayer player = EntityArgument.getPlayer(ctx, "player");
+        IKnowledgeProvider provider;
+        try {
+            provider = player.getCapability(PECapabilities.KNOWLEDGE_CAPABILITY).orElseThrow(() -> new IllegalStateException("Failed to get knowledge provider."));
+        } catch(IllegalStateException ignore) {
+            ctx.getSource().sendFailure(PELang.COMMAND_PROVIDER_FAIL.translateColored(ChatFormatting.RED));
+            return 0;
+        }
```
pupnewfster: Using exceptions as code flow is undesirable.
```java
Optional<IKnowledgeProvider> cap = player.getCapability(PECapabilities.KNOWLEDGE_CAPABILITY).resolve();
if (cap.isEmpty()) {
    ctx.getSource().sendFailure(PELang.COMMAND_PROVIDER_FAIL.translateColored(ChatFormatting.RED));
    return 0;
}
IKnowledgeProvider provider = cap.get();
```

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+                    }
+                }
+                case REMOVE -> {
+                    if (value.compareTo(BigInteger.ZERO) < 0) {
+                        action = ActionType.ADD;
+                        value = value.abs();
+                    }
+                }
+                case SET, TEST -> {
+                    if (value.compareTo(BigInteger.ZERO) < 0) value = null;
```
pupnewfster: No same line if statements, please use brackets. (Granted this if statement goes away after moving it around the switch)

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+
+        try {
+            value = new BigInteger(val);
+            switch (action) {
+                case ADD -> {
+                    if (value.compareTo(BigInteger.ZERO) < 0) {
+                        action = ActionType.REMOVE;
+                        value = value.abs();
+                    }
+                }
```
pupnewfster: This might be cleaner as:
`case ADD, REMOVE -> {`
and then add a helper method and have `action = action.invert();`

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+        if (action == ActionType.GET) {
+            ctx.getSource().sendSuccess(PELang.COMMAND_EMC_GET_SUCCESS.translate(player.getDisplayName(), formatEMC(provider.getEmc())), true);
+            return 1;
+        }
+        String val = StringArgumentType.getString(ctx, "value");
+        @Nullable BigInteger value = null;
+
+        try {
+            value = new BigInteger(val);
+            switch (action) {
```
pupnewfster: This entire switch statement can be surrounded by the `if (value.compareTo(BigInteger.ZERO) < 0) {`

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+        int response = 1;
+        BigInteger newEMC = provider.getEmc();
+        switch(action) {
+            case ADD -> {
+                newEMC = newEMC.add(value);
+                ctx.getSource().sendSuccess(PELang.COMMAND_EMC_ADD_SUCCESS.translate(formatEMC(value), player.getDisplayName(), formatEMC(newEMC)), true);
+            }
+            case REMOVE -> {
+                newEMC = newEMC.subtract(value);
+                ctx.getSource().sendSuccess(PELang.COMMAND_EMC_REMOVE_SUCCESS.translate(formatEMC(value), player.getDisplayName(), formatEMC(newEMC)), true);
```
pupnewfster: Why is a success being printed if it potentially will fail lower down due to being negative? It seems more reasonable to do that check up here (as none of the other cases can make it become negative anyway).

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+                ctx.getSource().sendSuccess(PELang.COMMAND_EMC_SET_SUCCESS.translate(player.getDisplayName(), formatEMC(value)), true);
+            }
+            case TEST -> {
+                boolean canTake = newEMC.compareTo(value) >= 0;
+                if(canTake) {
+                    ctx.getSource().sendSuccess(PELang.COMMAND_EMC_TEST_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), formatEMC(value)), true);
+                } else {
+                    ctx.getSource().sendFailure(PELang.COMMAND_EMC_TEST_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), formatEMC(value)));
+                    response = 0;
+                }
```
pupnewfster: Just return the response at the end of the TEST case. And then if statement lower down doesn't need to validate the action isn't test and also doesn't need to validate the response is equal to one.

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+                newEMC = newEMC.subtract(value);
+                ctx.getSource().sendSuccess(PELang.COMMAND_EMC_REMOVE_SUCCESS.translate(formatEMC(value), player.getDisplayName(), formatEMC(newEMC)), true);
+            }
+            case SET -> {
+                newEMC = value;
+                ctx.getSource().sendSuccess(PELang.COMMAND_EMC_SET_SUCCESS.translate(player.getDisplayName(), formatEMC(value)), true);
+            }
+            case TEST -> {
+                boolean canTake = newEMC.compareTo(value) >= 0;
+                if(canTake) {
```
pupnewfster: Inline the boolean into the if

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+            .then(Commands.literal("get")
+                .requires(PEPermissions.COMMAND_EMC_GET)
+                .then(Commands.argument("player", EntityArgument.player())
+                    .executes((ctx) -> handle(ctx, ActionType.GET))
+                )
+            );
+    }
+
+    private static MutableComponent formatEMC(BigInteger emc) {
+        return MutableComponent.create(TransmutationEMCFormatter.formatEMC(emc).getContents()).withStyle(ChatFormatting.GRAY);
```
pupnewfster: I believe `TextComponentUtil.build(ChatFormatting.GRAY, TransmutationEMCFormatter.formatEMC(emc))` should work, alternatively this could in theory use `TransmutationEMCFormatter.formatEMC(emc).copy().withStyle(ChatFormatting.GRAY)`

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+                    .executes((ctx) -> handle(ctx, ActionType.CLEAR))
+                )
+            )
+            .then(Commands.literal("learn")
+                .requires(PEPermissions.COMMAND_KNOWLEDGE_LEARN)
+                .then(Commands.argument("player", EntityArgument.player())
+                    .then(Commands.argument("item", ItemArgument.item(context))
+                        .executes((ctx) -> handle(ctx, ActionType.LEARN))
+                    )
+                )
```
pupnewfster: These could use a helper like mentioned in the other command.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private static int handle(CommandContext<CommandSourceStack> ctx, ActionType action) throws CommandSyntaxException {
+        ServerPlayer player = EntityArgument.getPlayer(ctx, "player");
+        IKnowledgeProvider provider;
+        try {
+            provider = player.getCapability(PECapabilities.KNOWLEDGE_CAPABILITY).orElseThrow(() -> new IllegalStateException("Failed to get knowledge provider."));
+        } catch (IllegalStateException ignore) {
+            ctx.getSource().sendFailure(PELang.COMMAND_PROVIDER_FAIL.translateColored(ChatFormatting.RED));
+            return 0;
+        }
```
pupnewfster: Same thing here as other command, exceptions as code flow are not ideal.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+        if (action == ActionType.CLEAR) {
+            if (provider.getKnowledge().size() == 0) {
+                ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_CLEAR_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName()));
+                return 0;
+            }
+            provider.clearKnowledge();
+            provider.sync(player);
+            ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_CLEAR_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName()), true);
+            return 1;
+        }
```
pupnewfster: This needs some evaluation as we have a `clearknowledge` command that seems to have more functionality than this

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,157 @@
...
+public class EMCCMD {
+    private enum ActionType {
+        ADD,
+        REMOVE,
+        SET,
+        GET,
+        TEST
+    }
+    public static ArgumentBuilder<CommandSourceStack, ?> register(CommandBuildContext context) {
+         return Commands.literal("emc")
```
pupnewfster: We should debate whether to name this `playeremc` to avoid potentially confusing people who try to set the emc value of an item.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+
+        ItemStack item = new ItemStack(ItemArgument.getItem(ctx, "item").getItem());
```
pupnewfster: Is there a reason this is using an `ItemStack` rather than just directly using `ItemInfo`? Given if we cared about NBT for the command we should be having the argument create the stack not doing it ourselves here.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+                    ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_UNLEARN_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+                }
+            }
+            case TEST -> {
+                if (provider.hasKnowledge(item)) {
+                    ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_TEST_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+                } else {
+                    ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_TEST_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+                    response = 0;
+                }
```
pupnewfster: Same as for other command, just return the response rather than having an if statement below just to do things differently for this case.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+            ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_INVALID.translateColored(ChatFormatting.RED, item.getDisplayName()));
+            return 0;
+        }
+
+        int response = 1;
+        switch (action) {
+            case LEARN -> {
+                if (provider.hasKnowledge(item)) {
+                    ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_LEARN_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+                    response = 0;
```
pupnewfster: Just fail immediately `return 0;`

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+                } else {
+                    provider.addKnowledge(item);
+                    provider.sync(player);
+                    ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_LEARN_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+                }
+            }
+            case UNLEARN -> {
+                if (!provider.hasKnowledge(item)) {
+                    ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_UNLEARN_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+                    response = 0;
```
pupnewfster: Just fail immediately `return 0;`

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,126 @@
...
+
+        int response = 1;
+        switch (action) {
+            case LEARN -> {
+                if (provider.hasKnowledge(item)) {
+                    ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_LEARN_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+                    response = 0;
+                } else {
+                    provider.addKnowledge(item);
+                    provider.sync(player);
```
pupnewfster: Why are these syncs here given the fact there is the syncKnowledgeChange below? I don't believe both should be necessary?

pupnewfster (changes requested):
Left some comments, also this project uses tabs over spaces.

pupnewfster:
Datagen task should just be `runData`. Which will also update the caches so that files don't have to be generated for other people who are running it in the future if nothing changes. It also updates en_ud automatically by running the datagen.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
DonovanDMC: We're creating a stack so we can display it in chat later via `ItemStack#getDisplayName`

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
DonovanDMC: Main differences I can see,

ProjectE:
* Succeeds regardless
* supports multiple players

This Version:
* Success is dependent on actual success of operation
* limited to one player

Ideally, it could be modified to support multiple players and return the success count

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,122 @@
...
+					provider.removeKnowledge(item);
+					provider.sync(player);
+					ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_UNLEARN_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+				}
+			}
+			case TEST -> {
+				if (provider.hasKnowledge(item)) {
+					ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_TEST_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+					return 1;
+				} else {
```
pupnewfster: Unneeded else, can just be implicit due to the fact the if statement always returns.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,122 @@
...
+					provider.addKnowledge(item);
+					provider.sync(player);
+					ctx.getSource().sendSuccess(PELang.COMMAND_KNOWLEDGE_LEARN_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), item.getDisplayName()), true);
+				}
+			}
+			case UNLEARN -> {
+				if (!provider.hasKnowledge(item)) {
+					ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_UNLEARN_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+					return 0;
+				} else {
```
pupnewfster: Unneeded else, can just be implicit due to the fact the if statement always returns.

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
```diff
@@ -0,0 +1,122 @@
...
+			ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_INVALID.translateColored(ChatFormatting.RED, item.getDisplayName()));
+			return 0;
+		}
+
+		switch (action) {
+			case LEARN -> {
+				if (provider.hasKnowledge(item)) {
+					ctx.getSource().sendFailure(PELang.COMMAND_KNOWLEDGE_LEARN_FAIL.translateColored(ChatFormatting.RED, player.getDisplayName(), item.getDisplayName()));
+					return 0;
+				} else {
```
pupnewfster: Unneeded else, can just be implicit due to the fact the if statement always returns.

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,137 @@
...
+			}
+			case SET -> {
+				newEMC = value;
+				ctx.getSource().sendSuccess(PELang.COMMAND_EMC_SET_SUCCESS.translate(player.getDisplayName(), formatEMC(value)), true);
+			}
+			case TEST -> {
+				if(newEMC.compareTo(value) >= 0) {
+					ctx.getSource().sendSuccess(PELang.COMMAND_EMC_TEST_SUCCESS.translateColored(ChatFormatting.GREEN, player.getDisplayName(), formatEMC(value)), true);
+					return 1;
+				} else {
```
pupnewfster: Unneeded else, can just be implicit due to the fact the if statement always returns.

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
```diff
@@ -0,0 +1,137 @@
...
+import moze_intel.projecte.utils.text.TextComponentUtil;
+import net.minecraft.ChatFormatting;
+import net.minecraft.commands.CommandBuildContext;
+import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.commands.Commands;
+import net.minecraft.commands.arguments.EntityArgument;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.server.level.ServerPlayer;
+
+import javax.annotation.Nullable;
```
pupnewfster: We like forge are using jetbrain's annotations for ProjectE

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
pupnewfster: Agreed, that ideally this would be modified to support multiple. Also the existing PE one should have a comment added `//TODO - 1.20: Remove, replaced be: x`

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
pupnewfster: This still needs and answer

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
pupnewfster: At which point is there a good reason we aren't allowing learning items with specific NBT (simplified by the persistent data)? by using the argument to create the stack?

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
DonovanDMC: image
This should be sufficient

src/main/java/moze_intel/projecte/network/commands/KnowledgeCMD.java
DonovanDMC: I'm pretty sure a stack argument doesn't exist in brigadier, and I have no idea how to go about using that `NSSItemArgument` class

Only way I can see it being usable currently is getting a string, which isn't what we want

src/main/java/moze_intel/projecte/network/commands/EMCCMD.java
DonovanDMC: Most confusion *should* be dissuaded by further commands requiring a player

If you wanted to get into it further, this command could be expanded to incorporate both items and users, say:
* emc
* * players
* * * get/set/test/add/remove
* * items
* * * get/set

or `itememc`/`playeremc`, as 4 levels deep can get quite confusing

`get` for items would be a bit wonky, and not have a truly practical use (we can't realistically return the value, as that caps out at the 32 bit integer limit, and nobody's going to be running the mess that would be `/projecte emc items get minecraft:oak_button`, they'll just use JEI or go into creative)

James103:
This should fix https://github.com/sinkillerj/ProjectE/issues/2024, as far as I can tell.

Also, will you please make sure that all instances of the term `EMC` are capitalized in the strings you add? Make sure to do this both in the generated lang file as well as the file that generates the lang file, and/or make sure to run the datagen task after changing the files. The rest of the strings have the `EMC` term capitalized.

PR #2253, opened 2022-11-10
schedule bonus ticks

wagyourtail:
schedule at the end of the tick, drop if lagging too much

add config option for max tps drop TimeWatch can cause

is this a good idea?

### DIFF
--- a/src/main/java/moze_intel/projecte/gameObjs/items/rings/TimeWatch.java
+++ b/src/main/java/moze_intel/projecte/gameObjs/items/rings/TimeWatch.java
@@ -183,8 +183,12 @@ public class TimeWatch extends PEToggleItem implements IPedestalItem, IItemCharg
 	public static void completeTick(Level level, BooleanSupplier haveTime) {
 		long currentTime = System.currentTimeMillis();
 		double maxTPSLost = ProjectEConfig.server.effects.maxTPSLoss.get();
-		double asMS = 1000d / (20 - maxTPSLost) - 50;
-
+		double asMS;
+		if (maxTPSLost >= 20 || maxTPSLost < 0) {
+			asMS = 10000;
+		} else {
+			asMS = 1000d / (20 - maxTPSLost) - 50;
+		}
 		// we now need to turn the 2 queues into a schedule that overshoots by the same ratio of ticks done to ticks scheduled for each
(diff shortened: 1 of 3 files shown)


pupnewfster:
This definitely sounds like something that would be good as a config option at the least, will try and take a look at it whenever I next get around to working on ProjectE

src/main/java/moze_intel/projecte/gameObjs/items/rings/TimeWatch.java
```diff
@@ -171,22 +160,101 @@ private void speedUpRandomTicks(Level level, int bonusTicks, AABB bBox) {
...
+				// we have run out of time, so we will drop the rest of the queue
+//				System.out.println("TimeWatch: Ran out of time, dropping " + (totalTicks - ticksTot) + " extra block ticks");
+				break;
+			}
+
+			// now we run for ticksleft / divisor ticks
+			for (SpedUpItem item : queue) {
+
+				// reasonable range so a single doesn't hog too much
+				int ticks = Mth.clamp(item.getTicksLeft() / divisor, 1, Math.min(totalTicks, 1_000));
```
wagyourtail, line 238: there's probably a better function usable here, 

the idea is to get a ratio so that things that have 100 bonus ticks don't get the same as things with only like 10. but we also don't want this while loop to take too long for a single iteration, since it only checks if it's out of time each loop

pupnewfster:
I am going to look into this in the next few days, though odds are I will more so pull it out of the PR/reimplement and copy bits as I am currently finishing up 1.19 before moving to 1.20 and this PR is targeted at 1.18

pupnewfster:
https://github.com/pupnewfster/ProjectE/commit/a90c2ac4be4a41b99f2f2258596e6c263dd82981 was my experimentation with this and there are some things I am not quite sure about in regards to things like assuming all blocks have same number of ticks given to them then they do one at a time a single extra tick, but it means validation has to be done again as if one of the blocks that got accelerated was a miner or something then the other block may no longer exist. So for now I am shelving this until 1.20.3+ when mojang is adding a tick command to make testing this easier and potentially if we are lucky maybe is doing some cleanup of how things are ticked to allow batching them easier? (doubtful but we can hope)

PR #2280, opened 2023-05-23
mc1.18.x expand covalence loss range and update JEI api

Phys09:
* Update jei api maven url for 1.18.x
JEI API server is down. According to mezz's discord server, we should be using blamejared instead.

* Increase range for covalenceLoss and update description
The main reason this PR was made in the first place.
Allows the "covalenceLoss" value to accept 0 and values past 1.

Accepting 0 lets people "delete" EMC from everything easily if they really want to play around with it.

Values above 1 do work, but is of course not balanced. Mainly added for people who want to mess around on a server.
You can change this back to 1.0 if you want.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -241,7 +241,8 @@ minecraft {
 repositories {
     maven {
         name 'JEI'
-        url "https://dvs1.progwml6.com/files/maven"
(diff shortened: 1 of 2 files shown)


sinkillerj:
This range is intentional, and there are no plans to change it.

Issue #2300, opened 2023-09-13
CraftTweaker integration WorldTransmutation.remove() is non-functional


loldev69:
**Tickets that do not conform to this template will be closed without comment**

Exact ProjectE version (do not say "latest", "latest on Curse", or similar): 1.18.2-PE1.0.1

Exact Forge version: tested on 40.2.0 and 40.2.10

Steps to reproduce:
install crafttweaker and run the following script
`import crafttweaker.api.block.Block;`
`import mods.projecte.WorldTransmutation;`
`WorldTransmutation.remove(<block:minecraft:gravel>.getDefaultState(), <block:minecraft:sandstone>.getDefaultState());`

What I expected to happen: 
the world transmutation from gravel to sandstone was removed

What happened instead:
the world transmutation from gravel to standstone was not removed, no error was logged and it seems that the script had no effect

Additional notes:
the other two WorldTransmutation functions seem to be working correctly, the following script does in fact remove all transmutations then adds back the gravel to sandstone transmutation.
`import crafttweaker.api.block.Block;`
`import mods.projecte.WorldTransmutation;`
`WorldTransmutation.removeAll();`
`WorldTransmutation.add(<block:minecraft:gravel>.getDefaultState(), <block:minecraft:sandstone>.getDefaultState());`

pupnewfster:
I see what is wrong and will fix it, a workaround for this should be to pass the state twice:
```zs
WorldTransmutation.remove(<block:minecraft:gravel>.getDefaultState(), <block:minecraft:sandstone>.getDefaultState(), <block:minecraft:sandstone>.getDefaultState());
```

PR #2358, opened 2024-08-26
Revert change removing the ability to use the Philosophers Stone's alt Function

Anoyomouse:
The philosopher stone's alt function is currently broken (you cannot hold down shift to change stone into grass for example)

By reverting this single change (from #2300) (made in Dec 2023), the alt function starts working again

Not sure if this breaks craft tweaker, but it definitely fixes the issue


pupnewfster:
Unsure about how sin does version numbers so if you want to revert the version bump I will merge this (though I am unsure if/when we will do another 1.20.1 release). In the meantime I have cherrypicked your commit to the WIP 1.21 branch https://github.com/sinkillerj/ProjectE/commit/aed536b64807dea5c6624cbc27ea2b628fd0b508

PR #2468, opened 2026-06-16
Rewrite japanese localization

twister716:
The Japanese translation of ProjectE was very old, and some parts were unsuitable, so I rewrote the translation from scratch.
I reused parts from the old translation that were relevant, and adjusted parts that needed to match the Minecraft translation.


sinkillerj:
Thank you for putting in the work on this, my own Japanese is fairly bad still, especially written, so that old translation has remained for quite some time. I will review this and get it merged in as soon as I have time, I see you are rather known so I don't suspect any issues preventing inclusion into the next released versions.

twister716:
Thank you. On this side, we are also checking the operation of the latest version of ProjectE by turning it into a resource pack, and confirming that there is no problem with the operation. Good luck with the development!

PR #2475, opened 2026-07-23
Fix EMC mapper correctness and failure isolation

raziel23x:
## Summary

This PR hardens the EMC mapper and the code paths that directly feed or consume mapped values. It fixes deterministic conversion identity, remap publication, forced-conversion liveness, recipe mapping, component-bearing item values, exact fractional intermediates, mapper ordering, and diagnostics.

The work is intentionally limited to EMC mapping and component value resolution. It does not add content, rebalance EMC, alter progression, or redesign unrelated ProjectE systems.

## Correctness fixes

- Include recipe output count and arithmetic strategy in conversion identity so mathematically different conversions cannot collapse into one set entry.
- Preserve the exact forced-conversion instance in reverse dependency tracking and remove stale dependencies when a forced conversion is replaced.
- Revalidate forced conversions after cycle and exploit cleanup so dead dependency chains cannot retain ghost EMC.
- Reject invalid negative fixed values while preserving the existing free-value sentinel where the API permits it.
- Defensively copy conversion ingredient maps.
- Keep recoverable exact fractions beyond the old `long` denominator limit, with a finite 256-bit denominator guard so descending fractional cycles still terminate.
- Enforce final positive-`long` publication boundaries.

## Remap lifecycle and failure isolation

- Keep the last known-good EMC map available until a complete replacement is ready.
- Abort a remap when an EMC mapper fails instead of continuing and publishing partial values.
- Prepare dependent component caches before swapping the live map and roll them back if preparation fails.
- Clear temporary `NSSFake` namespace and diagnostic-group state after every mapper and after `finishCollection`, including exceptional exits.

## Recipe and source determinism

- Preserve loader discovery order for equal-priority annotation-loaded mappers and avoid priority collisions between objects with equal `equals`/`hashCode` implementations.
- Apply custom conversion files in stable resource-ID order.
- Preserve CraftTweaker registration order, with the most recent repeated registration receiving final precedence.
- Give every recipe-mapper pass its own unhandled list so a mapper that handles nothing cannot contaminate later passes.
- Isolate malformed ingredient alternatives instead of allowing one broken stack to abort the entire remap.
- Fail malformed single ingredients and outputs closed.
- Subtract the full count and exact state of crafting remainders, with checked arithmetic.

## Component-bearing item safety

- Treat an exact component-bearing `ItemInfo` mapping as authoritative instead of applying component processors a second time.
- Continue using processors when only the component-less base item has a mapping.
- Calculate durability scaling without overflowing a valid final `long` result.
- Reject stacks at or beyond maximum damage.
- Reject negative stored EMC and overflowed stored-EMC addition.
- Fail a broken third-party component processor closed, stop later processors from observing a partial value, and log once per processor class between cache refreshes.

## Diagnostics

- Add an API-compatible default diagnostic hook for recipe mappers to explain intentionally unsupported dynamic/stateful recipes.
- Classify `minecraft:armor_dye` as an intentional dynamic skip because its output color is calculated from and preserves input components.
- Separate intentionally skipped recipes from unexpectedly unhandled recipes.
- Report profitable conversion mismatches using total output count and total output value.
- Distinguish fixed/custom retained values from forced-conversion mappings.
- Keep the existing `logFoundExploits` config key; only its user-facing description and diagnostic wording change.

## API compatibility

- No public method or type was removed.
- No existing public method signature changed.
- `IRecipeTypeMapper#getExpectedUnhandledReason` is an additive default method, preserving binary compatibility for existing implementations.
- Other API-source changes are JavaDoc clarifications describing existing zero-amount semantics, exact component mapping authority, processor failure behavior, and equal-priority ordering.

## Build support

- The optional EMI development dependency remains version `1.1.20` for Minecraft `1.21.1`.
- EMI resolves from its official Sleeping Town Maven.
- Normal compilation uses `dev.emi:emi-neoforge:1.1.20+1.21.1:api`; datagen and local runtime use the full artifact.
- No runtime dependency is bundled or added to ProjectE.

## Validation

Validated from the preserved review-response tree after applying the current review round:

- `gradlew.bat clean runData` - successful
- `gradlew.bat clean test` - successful: 149 tests, 0 failures, 0 errors
- `gradlew.bat clean build` - successful
- `gradlew.bat runClient` - successful
- Existing `TEST` world loaded successfully
- `/reload` completed successfully, including CraftTweaker reload
- Spruce log EMC remained 32 before and after reload
- 1,466 EMC values registered at startup and after reload
- Startup mapping: 155 ms
- Reload mapping: 46 ms
- Crafting recipes: 1,048 handled, 1 intentionally skipped, 0 unexpectedly unhandled, 1,049 total
- All other supported recipe types: 0 unexpectedly unhandled
- CraftTweaker log: 0 warnings, 0 errors
- Runtime logs: 0 errors, 0 exceptions
- `git diff --check` - successful
- Final diff: 41 files total - 26 production/API/build/generated files and 15 test files

The profitable-conversion mismatch messages observed during validation are the expected diagnostics controlled by the existing mapping diagnostic option; they are not mapper failures. The final runtime retained the existing 1,466-value baseline. No gameplay tick system or recurring scan was added.

## Review response: pupnewfster

- Switched EMI development resolution to the official Sleeping Town Maven. Normal compilation uses EMI's `:api` classifier; datagen and local runtime use the full artifact.
- Kept deterministic CraftTweaker insertion order and clarified why repeated registrations are removed and reinserted: replacing a linked-map value retains the original position, while later script actions must retain later precedence when overlapping registrations occur between them. Identical repeats remain permitted because a later action may intentionally reassert precedence.
- Restored the armor-dye TODO while retaining the explanation of its current dynamic-skip classification.
- Replaced the class-level armor-dye check with a direct `instanceof` check on the recipe value and updated the tests to use real `RecipeHolder` values.
- Removed the unnecessary explicit `static` qualifier from `PrioritizedElement`.
- Applied the suggested stream-based priority sorting implementation.
- Ran `runData` and included the updated generated `en_us`, `en_ud`, and datagen cache files.
- Strengthened conversion collections to `SequencedSet` and documented why remove/reinsert gives a newer forced conversion later precedence.
- Added a concrete CraftTweaker regression test for reasserting an unchanged specific-item value after an intervening broad-tag assignment.
- Narrowed malformed recipe-stack isolation to `IllegalArgumentException`; unexpected runtime failures now escape, with regression coverage.
- Applied the requested test annotations, collection cleanup, comparator, reflection helper, test setup, comments, and constant usage changes.
- Benchmarked the denominator guard with an adversarial 300-link descending-fraction chain: median graph generation was 0.172 ms at 63 bits and 0.535 ms at 256 bits on Java 21.0.11. This diagnostic microbenchmark is not JMH or representative of a normal ProjectE graph.

## Compatibility and impact

- **Gameplay/balance:** No intended balance or progression changes. Existing development-instance EMC values remained unchanged.
- **Save data:** No save-format or migration changes.
- **Networking:** No packet-format change.
- **Public API:** Additive default method plus documentation clarifications; no removals.
- **Dependencies:** Development-only EMI resolution uses the official Sleeping Town Maven; normal compilation uses the API classifier, and the EMI version remains unchanged.
- **Performance:** No gameplay tick-path additions. Startup/reload timings remained within the previously observed development-instance range. An adversarial denominator-bound microbenchmark showed a 3.10x median relative cost for 256 bits versus 63 bits, but only 0.535 ms versus 0.172 ms for the complete 300-link diagnostic graph.

## Reviewer attention

The behavior changes most likely to deserve focused review are:

1. A mapper exception now aborts the remap and preserves the prior map instead of publishing partial data.
2. Exact component-bearing mappings bypass component processors by design.
3. Conflicting custom conversion files use resource-ID precedence, and CraftTweaker uses script registration precedence.
4. Exact fractions use a finite 256-bit denominator guard rather than the former `Long.MAX_VALUE` denominator check.
5. EMI resolves from the official Sleeping Town Maven; normal compilation uses the API classifier, while datagen and local runtime use the full artifact.


build.gradle
```diff
@@ -250,8 +250,9 @@ dependencies {
         localRuntime("mezz.jei:jei-${minecraft_version}-neoforge:${jei_version}")
     }
 
-    compileOnly("dev.emi:emi-neoforge:${emi_version}+${minecraft_version}:api")
-    def emi = "dev.emi:emi-neoforge:${emi_version}+${minecraft_version}"
+    // Modrinth exposes the pinned EMI build by immutable project and version IDs.
+    def emi = "maven.modrinth:${emi_modrinth_project_id}:${emi_modrinth_version_id}"
+    compileOnly(emi)
```
pupnewfster: This is changing it from compiling against just the API to compiling against the whole the jar. Also even if we do decide to do this change, I would rather pull it from cursemaven like we do for things like jade, rather than pull it from modrinth.

src/main/java/moze_intel/projecte/integration/crafttweaker/mappers/CrTCustomEMCMapper.java
```diff
@@ -18,9 +18,12 @@
 @EMCMapper(requiredMods = "crafttweaker")
 public class CrTCustomEMCMapper implements IEMCMapper<NormalizedSimpleStack, Long> {
 
-	private static final Object2LongMap<NormalizedSimpleStack> customEmcValues = new Object2LongOpenHashMap<>();
+	//CraftTweaker actions are ordered. Preserve that order so overlapping tag and item assignments have deterministic, script-defined precedence.
+	private static final Object2LongMap<NormalizedSimpleStack> customEmcValues = new Object2LongLinkedOpenHashMap<>();
 
 	public static void registerCustomEMC(@NotNull NormalizedSimpleStack stack, long emcValue) {
+		//Move repeat registrations to the end so the most recent script action also has the latest expansion precedence.
+		customEmcValues.removeLong(stack);
```
pupnewfster: Realistically I don't think we really need to worry about repeat entries? Like sure I get changing it to a linked map from not linked so that the order is deterministic for tags vs non-tags. I would not be opposed to throwing an exception if it already has an identical value registered though, as that is likely to be a copy-paste error in the script.

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
```diff
@@ -50,7 +54,7 @@ public final boolean handleRecipe(IMappingCollector<NormalizedSimpleStack, Long>
 			} else if (recipe instanceof FireworkRocketRecipe) {
 				return MappingConfig.isEnabled(FireworkProcessor.INSTANCE);
 			}
-			//TODO: Do we eventually want to try and figure out how to handle the armor dye recipe?
+			//Armor dye is intentionally left unmapped and classified separately because its output is dynamic.
```
pupnewfster: You can leave the comment you added as it is a decent explanation, but the TODO still is worthwhile, as we do handle some other dynamic things via the DataComponentProcessors

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
```diff
@@ -66,6 +70,17 @@ public final boolean canHandle(RecipeType<?> recipeType) {
 		return recipeType == RecipeType.CRAFTING;
 	}
 
+	@Override
+	@Nullable
+	public final String getExpectedUnhandledReason(RecipeHolder<?> recipeHolder, RegistryAccess registryAccess) {
+		return getExpectedUnhandledReason(recipeHolder.value().getClass());
```
pupnewfster: Why is this going to the class level, instead of just doing an instanceof on the value?

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
```diff
@@ -100,8 +93,20 @@ public static List<IEMCMapper<NormalizedSimpleStack, Long>> getEMCMappers() {
...
+		elements.sort(Comparator.comparingInt((PrioritizedElement<T> element) -> element.priority()).reversed()
+				.thenComparingInt(PrioritizedElement::discoveryOrder));
+		List<T> sorted = new ArrayList<>(elements.size());
+		for (PrioritizedElement<T> element : elements) {
+			sorted.add(element.value());
+		}
+		return sorted;
+	}
+
+	static record PrioritizedElement<T>(T value, int priority, int discoveryOrder) {
```
pupnewfster: Unnecessary `static` qualifier

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
```diff
@@ -100,8 +93,20 @@ public static List<IEMCMapper<NormalizedSimpleStack, Long>> getEMCMappers() {
...
+	}
+
+	static <T> List<T> sortByPriority(List<PrioritizedElement<T>> elements) {
+		elements.sort(Comparator.comparingInt((PrioritizedElement<T> element) -> element.priority()).reversed()
+				.thenComparingInt(PrioritizedElement::discoveryOrder));
+		List<T> sorted = new ArrayList<>(elements.size());
+		for (PrioritizedElement<T> element : elements) {
+			sorted.add(element.value());
+		}
+		return sorted;
```
pupnewfster: I believe this should work as well, and is a bit clearer about the operations being performed to translate the list
```suggestion
		return elements.stream()
				.sorted(Comparator.<PrioritizedElement<T>>comparingInt(PrioritizedElement::priority)
						.reversed()
						.thenComparingInt(PrioritizedElement::discoveryOrder)
				).map(PrioritizedElement::value)
				.toList();
```

src/main/java/moze_intel/projecte/config/PEConfigTranslations.java
```diff
@@ -115,9 +115,9 @@ public enum PEConfigTranslations implements IConfigTranslation {
 	MAPPING_PREGENERATED("mapping.pregenerated", "Pregenerate EMC",
 			"When the next EMC mapping occurs write the results to config/ProjectE/pregenerated_emc.json and only ever run the mapping again when that file "
 			+ "does not exist, this setting is set to false, or an error occurred parsing that file."),
-	MAPPING_LOG_EXPLOITS("mapping.log_exploits", "Log Known EMC Exploits",
-			"Logs known EMC Exploits. This can not and will not find all possible exploits. This will only find exploits that result in fixed/custom emc values "
-			+ "that the algorithm did not overwrite. Exploits that derive from conversions that are unknown to ProjectE will not be found."),
+	MAPPING_LOG_EXPLOITS("mapping.log_exploits", "Log Profitable EMC Conversion Mismatches",
```
pupnewfster, line 118: Please run `runData`, to update the generated lang files for the changed descriptions.

pupnewfster (changes requested):

build.gradle
raziel23x: Updated. EMI now resolves from CurseMaven using the pinned file ID `6205506`, consistent with the existing Curse dependency pattern. CurseForge does not publish EMI's API classifier separately, so both compile-only scopes use the pinned full jar. It remains development-only and is not bundled.

src/main/java/moze_intel/projecte/integration/crafttweaker/mappers/CrTCustomEMCMapper.java
raziel23x: I kept the linked map and clarified the reason for removing and reinserting repeats. Replacing a value in a linked map retains the key's original position, which would make a later script action execute with its old precedence when overlapping tag or item registrations were added between the two actions. I left identical repeats permitted because a later action may intentionally reassert precedence even when the value is unchanged.

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
raziel23x: Restored the TODO and kept the explanation of why armor dye is currently classified as an intentional dynamic skip.

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
raziel23x: Changed this to a direct `recipeHolder.value() instanceof ArmorDyeRecipe` check. The regression tests now use actual `RecipeHolder` values rather than passing recipe classes.

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
raziel23x: Removed the unnecessary explicit `static` qualifier.

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
raziel23x: Applied the suggested stream-based implementation.

src/main/java/moze_intel/projecte/config/PEConfigTranslations.java
raziel23x: Ran `clean runData` and included the updated generated `en_us.json`, `en_ud.json`, and datagen cache file. The resulting tree also passed `clean test` (147 tests), `clean build`, and `runClient`; the TEST world and `/reload` retained 1,466 EMC values and spruce-log EMC 32.

build.gradle
```diff
@@ -227,7 +227,6 @@ repositories { RepositoryHandler handler ->
         filter.includeGroup('mezz.jei')
         filter.includeGroup('org.openzen.zencode')
     })
-    exclusiveRepo(handler, 'https://maven.terraformersmc.com/', 'dev.emi')
```
pupnewfster, line 230: Turns out EMI switched mavens. Copy of a message from their announcement channel today:
> Unfortunately, time has not been kind to us, and the TerraformersMC Maven is in its final moments, to be shutting down indefinitely. My mods have been moved to the Sleeping Town maven (cross your fingers that we will not need to move again) and that means you'll need to update all of your build files that depend on EMI. I'm sorry, I now this sucks. Here's an easy block to copy. Other mods that use the TerraformersMC Maven will be in other locations, you'll have to follow up with other devs.
```gradle
repositories {
    maven {
        name = "Sleeping Town"
        url = "https://repo.sleeping.town/"
    }
}
```

So I guess this line can instead be changed to that maven, and then theoretically the rest of it might be able to be left as is?

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
```diff
@@ -30,10 +31,13 @@
 import net.minecraft.world.item.crafting.RecipeType;
 import net.minecraft.world.item.crafting.RepairItemRecipe;
 import net.minecraft.world.item.crafting.ShieldDecorationRecipe;
+import org.jetbrains.annotations.Nullable;
 
 @RecipeTypeMapper
 public class SpecialRecipeMarkHandledMapper implements IRecipeTypeMapper {
 
+	static final String ARMOR_DYE_SKIP_REASON = "Output color is calculated from and preserves the input item's data components.";
```
pupnewfster, line 41: `@VisibleForTesting`

src/main/java/moze_intel/projecte/integration/crafttweaker/mappers/CrTCustomEMCMapper.java
pupnewfster: What case would they be intentionally doing so though?

src/main/java/moze_intel/projecte/emc/arithmetic/FullBigFractionArithmetic.java
```diff
@@ -80,13 +82,10 @@ public BigFraction div(BigFraction a, long b) {
 			return getZero();
 		}
 		BigFraction result = a.divide(b);
-		//TODO: I believe checking the numerator is not necessary as we will clamp it further down the line
-		if (/*MathUtils.isGreaterThanLong(result.getNumerator()) ||*/ MathUtils.isGreaterThanLong(result.getDenominator())) {
```
pupnewfster, line 84: You can remove the now unused `MathUtils#isGreaterThanLong`

src/main/java/moze_intel/projecte/emc/arithmetic/FullBigFractionArithmetic.java
```diff
@@ -3,11 +3,13 @@
 import java.math.BigInteger;
 import moze_intel.projecte.api.ProjectEAPI;
 import moze_intel.projecte.api.mapper.arithmetic.IValueArithmetic;
-import moze_intel.projecte.utils.MathUtils;
 import org.apache.commons.math3.fraction.BigFraction;
 
 public class FullBigFractionArithmetic implements IValueArithmetic<BigFraction> {
 
+	//Keep exact intermediate fractions beyond long range, but bound descending cycles so BigInteger denominators cannot grow forever.
+	private static final int MAX_EXACT_DENOMINATOR_BITS = 256;
```
pupnewfster, line 11: This is quite a bit further than the old max of `63`, how much of a performance impact does doing this have?

src/main/java/moze_intel/projecte/emc/collector/MappingCollector.java
```diff
@@ -67,6 +67,15 @@ private void addConversionToIngredientUsages(Conversion conversion) {
 		}
 	}
 
+	private void replaceConversionInIngredientUsages(Conversion conversion) {
+		//Ensure the exact overwrite instance is retained if an equivalent regular conversion was already registered.
+		for (T ingredient : conversion.ingredientsWithAmount.keySet()) {
+			Set<Conversion> usesForIngredient = getUsesFor(ingredient);
```
pupnewfster: Should we strengthen the set type to a `SequencedSet`? Also please add a comment that we have to remove it as we want it to get the new order to have higher precedence

src/main/java/moze_intel/projecte/emc/components/DataComponentManager.java
```diff
@@ -55,36 +61,43 @@ static ItemInfo getPersistentInfo(@NotNull ItemInfo info) {
 
 	@Range(from = 0, to = Long.MAX_VALUE)
 	public static long getEmcValue(@NotNull ItemInfo info) {
-		//TODO: Fix this, as it does not catch the edge case that we have an exact match and then there are random added Data Components on top of it
-		// but that can be thought about more once we have the first pass complete. For example if someone put an enchantment on a potion
```
pupnewfster, line 59: I suspect this TODO is still relevant, as the issue is potion items have components that determine their variant, and thus are not part of the default components of just the item.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -199,10 +237,21 @@ private boolean addIngredient(Object2IntMap<NormalizedSimpleStack> ingredientMap
...
+		NormalizedSimpleStack normalizedStack = normalizeStack(stack, recipeID, "ingredient");
+		if (normalizedStack == null) {
+			return true;
+		}
+		ingredientMap.mergeInt(normalizedStack, 1, Constants.INT_SUM);
 		return false;
 	}
 
+	static <TYPE> void subtractCraftingRemainder(Object2IntMap<TYPE> ingredientMap, TYPE remainder, int count) {
+		if (count <= 0) {
```
pupnewfster: It is impossible for this to be negative. `ItemStack#getCount` clamps the returned value at zero

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -199,10 +237,21 @@ private boolean addIngredient(Object2IntMap<NormalizedSimpleStack> ingredientMap
...
+			throw new IllegalArgumentException("Crafting remainder count must be positive");
+		}
+		ingredientMap.mergeInt(remainder, Math.negateExact(count), Math::addExact);
```
pupnewfster: This `negateExact` is unnecessary, it specifies it only has issues for `Integer.MIN_VALUE`, but we already know that `count` is a non-negative value

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -134,17 +124,60 @@ public boolean handleRecipe(IMappingCollector<NormalizedSimpleStack, Long> mappe
...
+	}
+
+	@Nullable
+	static NormalizedSimpleStack normalizeStack(ItemStack stack, ResourceLocation recipeID, String stackRole,
+			Function<ItemStack, NormalizedSimpleStack> normalizer) {
+		try {
+			return normalizer.apply(stack);
+		} catch (RuntimeException e) {
+			ResourceLocation itemName = BuiltInRegistries.ITEM.getKey(stack.getItem());
+			PECore.LOGGER.error(LogUtils.FATAL_MARKER, "Error mapping recipe {}. Failed to normalize the {} stack for item {} ({}). "
```
pupnewfster, line 179: What cases can this actually fail?

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -134,17 +124,60 @@ public boolean handleRecipe(IMappingCollector<NormalizedSimpleStack, Long> mappe
...
+		mapper.addConversion(recipeOutput.getCount(), normalizedOutput, ingredientMap);
 		return true;
 	}
 
+	static List<ItemStack> getNormalizableMatches(ItemStack[] matches, ResourceLocation recipeID,
+			Object2IntMap<NormalizedSimpleStack> rawNSSMatches) {
+		return getNormalizableMatches(matches, recipeID, rawNSSMatches, NSSItem::createItem);
+	}
+
+	static List<ItemStack> getNormalizableMatches(ItemStack[] matches, ResourceLocation recipeID,
```
pupnewfster, line 150: `@VisibleForTesting`

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -134,17 +124,60 @@ public boolean handleRecipe(IMappingCollector<NormalizedSimpleStack, Long> mappe
...
+		NormalizedSimpleStack normalizedOutput = normalizeStack(recipeOutput, recipeID, "recipe output");
+		if (normalizedOutput == null) {
+			//The recipe is malformed and later generic mappers would encounter the same invalid output.
+			return true;
+		}
+		mapper.addConversion(recipeOutput.getCount(), normalizedOutput, ingredientMap);
 		return true;
 	}
 
+	static List<ItemStack> getNormalizableMatches(ItemStack[] matches, ResourceLocation recipeID,
```
pupnewfster, line 144: `@VisibleForTesting`

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
```diff
@@ -134,17 +124,60 @@ public boolean handleRecipe(IMappingCollector<NormalizedSimpleStack, Long> mappe
 							return true;
 						}
 					}
-				} else if (addIngredient(ingredientMap, stacks.getFirst(), recipeID)) {//There is only actually one non-empty ingredient
-					//Failed to add ingredient, bail but mark that we handled it as there is a 99% chance a later
-					// mapper would fail as well due to it being an invalid recipe
+				} else if (addIngredient(ingredientMap, stacks.getFirst(), recipeID)) {
+					//There is only one valid alternative. Failed to add it, so later generic mappers would encounter the same invalid state.
```
pupnewfster, line 129: Please keep the part of the comment about there being just an extremely high chance of a later one would also fail. Given technically if at some point we make `addIngredient` protected or something, it might not need remainders etc and thus another might not fail

src/main/java/moze_intel/projecte/emc/mappers/customConversions/CustomConversionMapper.java
```diff
@@ -58,40 +59,52 @@ public String getDescription() {
...
 		}
-		NSSFake.resetNamespace();
 		return loading;
 	}
 
+	static void addMappingsFromFiles(Map<ResourceLocation, CustomConversionFile> files, IMappingCollector<NormalizedSimpleStack, Long> mapper) {
+		//FileToIdConverter returns a Map without an ordering contract. Apply files by resource ID so conflicting fixed values and forced
+		//conversions have stable precedence that cannot change when an unrelated file changes the backing map's iteration order.
+		List<ResourceLocation> orderedIds = new ArrayList<>(files.keySet());
+		orderedIds.sort(null);
```
pupnewfster: This should use `ResourceLocation::compareNamespaced`

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+		@Override
+		public boolean equals(Object obj) {
+			return obj instanceof EqualElement;
+		}
+
+		@Override
+		public int hashCode() {
+			return 0;
+		}
```
pupnewfster: This feels super ugly, I get why you did it this way, but just make it a normal class instead of a record

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
```diff
@@ -100,8 +93,19 @@ public static List<IEMCMapper<NormalizedSimpleStack, Long>> getEMCMappers() {
...
+	static <T> List<T> sortByPriority(List<PrioritizedElement<T>> elements) {
+		return elements.stream()
+				.sorted(Comparator.<PrioritizedElement<T>>comparingInt(PrioritizedElement::priority)
+						.reversed()
+						.thenComparingInt(PrioritizedElement::discoveryOrder)
+				).map(PrioritizedElement::value)
+				.toList();
+	}
+
+	record PrioritizedElement<T>(T value, int priority, int discoveryOrder) {
```
pupnewfster, line 111: `@VisibleForTesting`

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
```diff
@@ -100,8 +93,19 @@ public static List<IEMCMapper<NormalizedSimpleStack, Long>> getEMCMappers() {
 				}
 			}
 		}
-		emcMappers.sort(Comparator.comparingInt(priorities::getInt).reversed());
-		return emcMappers;
+		return sortByPriority(emcMappers);
+	}
+
+	static <T> List<T> sortByPriority(List<PrioritizedElement<T>> elements) {
```
pupnewfster, line 101: `@VisibleForTesting`

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+@DisplayName("Test annotation-loaded element ordering")
+class AnnotationHelperTest {
+
+	@Test
+	@DisplayName("Higher priorities run first and equal priorities retain discovery order")
+	void testPriorityAndDiscoveryOrder() {
+		Object first = new Object();
+		Object second = new Object();
+		Object highPriority = new Object();
+		List<PrioritizedElement<Object>> elements = new ArrayList<>(List.of(
```
pupnewfster: You don't make use of the mutability anywhere, just use `List.of`, no need to wrap it in a `new ArrayList`

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+		Assertions.assertSame(first, sorted.get(1));
+		Assertions.assertSame(second, sorted.get(2));
+	}
+
+	@Test
+	@DisplayName("Equal mapper objects retain their own priorities")
+	void testEqualObjectsDoNotCollide() {
+		EqualElement lowPriority = new EqualElement("low");
+		EqualElement highPriority = new EqualElement("high");
+		List<PrioritizedElement<EqualElement>> elements = new ArrayList<>(List.of(
```
pupnewfster: You don't make use of the mutability anywhere, just use `List.of`, no need to wrap it in a `new ArrayList`

src/test/java/moze_intel/projecte/emc/SimpleGraphMapperDiagnosticsTest.java
```diff
@@ -0,0 +1,53 @@
...
+import moze_intel.projecte.emc.arithmetic.HiddenBigFractionArithmetic;
+import org.apache.commons.math3.fraction.BigFraction;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("Test graph mapper conversion diagnostics")
+class SimpleGraphMapperDiagnosticsTest {
+
+	private final IValueArithmetic<BigFraction> arithmetic = new HiddenBigFractionArithmetic();
```
pupnewfster: Is there a reason that this isn't behaving like `GraphMapperTest`, and doing this via:
```java
@BeforeEach
void setup() {
```

src/test/java/moze_intel/projecte/emc/SimpleGraphMapperDiagnosticsTest.java
```diff
@@ -0,0 +1,53 @@
...
+		Assertions.assertEquals(new BigFraction(outputValue), comparison.outputValue());
+		Assertions.assertEquals(outputCount, comparison.outputCount());
+		Assertions.assertEquals(new BigFraction(expectedTotalOutputValue), comparison.totalOutputValue());
+	}
+
+	@Test
+	@DisplayName("Test invalid output counts are rejected by diagnostics")
+	void testInvalidOutputCount() {
+		Assertions.assertThrows(IllegalArgumentException.class, () -> SimpleGraphMapper.compareConversionValues(
+				arithmetic, new BigFraction(1), new BigFraction(1), 0
```
pupnewfster: Use `BigFraction.ONE`

src/test/java/moze_intel/projecte/emc/HiddenFractionSpecificTest.java
```diff
@@ -105,6 +105,43 @@ void moltenEnderpearlWithConversionArithmetic() {
...
+		Assertions.assertEquals(0, values.getLong("subunit3"));
+		Assertions.assertEquals(1, values.getLong("recovered3"));
+	}
+
+
+	@Test
+	@DisplayName("Test exact fraction arithmetic retains a finite denominator bound")
+	void fullFractionArithmeticHasFiniteDenominatorBound() {
+		FullBigFractionArithmetic arithmetic = new FullBigFractionArithmetic();
+		BigFraction value = new BigFraction(1);
```
pupnewfster: Use `BigFraction.ONE`

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
```diff
@@ -0,0 +1,253 @@
...
+		ItemInfo container = createContainer(new ItemStack(Items.STONE, 2));
+		Assertions.assertEquals(0, DataComponentManager.getEmcValue(container),
+				"An unrepresentable nested value must fail closed rather than wrap or publish partial EMC");
+	}
+
+	@SuppressWarnings("unchecked")
+	private List<IDataComponentProcessor> mutableProcessors() throws ReflectiveOperationException {
+		Field field = DataComponentManager.class.getDeclaredField("processors");
+		field.setAccessible(true);
+		return (List<IDataComponentProcessor>) field.get(null);
```
pupnewfster:
```suggestion
		return ObfuscationReflectionHelper.getPrivateValue(DataComponentManager.class, null, "procesors");
```

(You also will be able to remove the `throws` declaration and the supress warnings, though given it is only used in one place you may want to then inline it)

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+		List<IDataComponentProcessor> processors = mutableProcessors();
+		processors.addFirst(later);
+		processors.addFirst(failing);
+		try {
+			Assertions.assertEquals(0, DataComponentManager.getEmcValue(damaged),
+					"An unexpected processor exception must fail the complete item value closed");
+			Assertions.assertFalse(laterProcessorRan.get(),
+					"Processors after a failure must not observe or publish a partially calculated value");
+		} finally {
```
pupnewfster, line 134: This needs a comment relating to the fact that the reason this finally block is even necessary is because we are modifying the backing processor list so then it might break other tests

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	private List<IDataComponentProcessor> mutableProcessors() throws ReflectiveOperationException {
+		Field field = DataComponentManager.class.getDeclaredField("processors");
+		field.setAccessible(true);
+		return (List<IDataComponentProcessor>) field.get(null);
+	}
+
+	private abstract static class TestProcessor implements IDataComponentProcessor {
```
pupnewfster, line 229: Put this private class below the helper methods that are currently below this class

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
```diff
@@ -0,0 +1,253 @@
...
+	@Test
+	@DisplayName("Nested container contents compose without double counting")
+	void testNestedContainerContentsCompose() {
+		Object2LongMap<ItemInfo> values = new Object2LongOpenHashMap<>();
+		values.put(ItemInfo.fromItem(Items.SHULKER_BOX), 100);
+		values.put(ItemInfo.fromItem(Items.STONE), 4);
+		EMCMappingHandler.updateEmcValues(values);
+
+		ItemStack inner = createContainerStack(new ItemStack(Items.STONE, 2));
+		ItemInfo outer = createContainer(inner);
```
pupnewfster, line 176: This should have a comment saying that while it is not possible to nest shulker boxes in vanilla, that doesn't preclude the possibility of a modded backpack allowing nesting things that may have more complex emc values

pupnewfster (changes requested):
I still have a handful of files to work through and review, but am leaving my current comments, and also noting that the EMC value of firework rockets is different with this PR vs without it

src/main/java/moze_intel/projecte/integration/crafttweaker/mappers/CrTCustomEMCMapper.java
raziel23x: Addressed in f4d387c1. The intentional identical-repeat case is: set a specific item to 100, apply a broad tag at 200 that includes it, then reassert the same specific item at 100 so the specific override has latest precedence again. I added a regression test proving the final order is broad tag, then specific item, with the specific value restored.

build.gradle
raziel23x: Updated in f4d387c1. EMI now resolves from the official Sleeping Town Maven. Normal compilation uses the `:api` classifier, while datagen and local runtime retain the full artifact. The compile and datagen dependency trees resolved successfully, and the full test/build validation passed.

src/main/java/moze_intel/projecte/emc/mappers/recipe/special/SpecialRecipeMarkHandledMapper.java
raziel23x: Added `@VisibleForTesting` to `ARMOR_DYE_SKIP_REASON` in f4d387c1.

src/main/java/moze_intel/projecte/emc/arithmetic/FullBigFractionArithmetic.java
raziel23x: Removed the now-unused `MathUtils#isGreaterThanLong` helper in f4d387c1.

src/main/java/moze_intel/projecte/emc/arithmetic/FullBigFractionArithmetic.java
raziel23x: I measured this before keeping the bound in f4d387c1. On Java 21.0.11, an intentionally adversarial 300-link descending-fraction chain had median `generateValues` time of 0.172 ms at 63 bits and 0.535 ms at 256 bits (3.10x); p95 was 0.241 ms versus 0.910 ms. A 20,000-iteration arithmetic-only loop was about 4.41x slower. This is a diagnostic microbenchmark rather than JMH or a representative normal ProjectE graph, but it confirms a meaningful relative cost with a small absolute cost in the adversarial graph. I kept 256 because the regression coverage requires exact fractions to recover after crossing the old `long` denominator range while still retaining a finite termination bound.

src/main/java/moze_intel/projecte/emc/collector/MappingCollector.java
raziel23x: Done in f4d387c1. The conversion collections now use `SequencedSet`, and the replacement path documents that removal before re-adding moves the newer forced conversion to the end so it receives the intended precedence.

src/main/java/moze_intel/projecte/emc/components/DataComponentManager.java
raziel23x: Restored the TODO in f4d387c1, specifically noting the unresolved case of extra components layered onto an exact component-bearing variant mapping.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Updated in f4d387c1. The helper now rejects only zero, and the impossible negative-count test was removed.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Removed the unnecessary `Math.negateExact` in f4d387c1; the validated nonzero count is now negated directly while `Math::addExact` still protects accumulation overflow.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Investigated and narrowed in f4d387c1. `NSSItem` can reject malformed identities such as empty/default stacks, unbound direct holders, and unregistered items with `IllegalArgumentException`. The isolation now catches only that exception. Unexpected runtime/programming failures escape and abort the remap, with regression coverage proving an `IllegalStateException` is not hidden.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Added `@VisibleForTesting` to this `getNormalizableMatches` overload in f4d387c1.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Added `@VisibleForTesting` to the injectable-normalizer overload in f4d387c1.

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
raziel23x: Restored that qualification in f4d387c1; the relevant comments now say there is an extremely high chance a later mapper would also fail rather than presenting it as absolute.

src/main/java/moze_intel/projecte/emc/mappers/customConversions/CustomConversionMapper.java
raziel23x: Changed the ordering comparator to `ResourceLocation::compareNamespaced` in f4d387c1.

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
raziel23x: Changed the test element from a record to a normal class in f4d387c1.

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
raziel23x: Added `@VisibleForTesting` to `sortByPriority` in f4d387c1.

src/main/java/moze_intel/projecte/utils/AnnotationHelper.java
raziel23x: Added `@VisibleForTesting` to `PrioritizedElement` in f4d387c1.

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
raziel23x: Replaced the unnecessary mutable wrapper with `List.of` in f4d387c1.

src/test/java/moze_intel/projecte/utils/AnnotationHelperTest.java
raziel23x: Replaced this unnecessary mutable wrapper with `List.of` in f4d387c1 as well.

src/test/java/moze_intel/projecte/emc/SimpleGraphMapperDiagnosticsTest.java
raziel23x: Changed the mapper initialization to a `@BeforeEach` setup method in f4d387c1, matching `GraphMapperTest`.

src/test/java/moze_intel/projecte/emc/SimpleGraphMapperDiagnosticsTest.java
raziel23x: Changed this to `BigFraction.ONE` in f4d387c1.

src/test/java/moze_intel/projecte/emc/HiddenFractionSpecificTest.java
raziel23x: Changed this to `BigFraction.ONE` in f4d387c1.

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
raziel23x: Applied this in f4d387c1. The test now uses `ObfuscationReflectionHelper.getPrivateValue`, and the reflective checked-exception/suppression plumbing was removed.

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
raziel23x: Added the requested explanation in f4d387c1: the test mutates the shared backing processor list, so the `finally` restoration is required to prevent contamination of later tests.

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
raziel23x: Moved the private test processor class below the helper methods in f4d387c1.

src/test/java/moze_intel/projecte/emc/components/DataComponentManagerTest.java
raziel23x: Added the requested comment in f4d387c1, noting that vanilla shulker nesting is prohibited but modded backpacks or other containers can still permit nested component-bearing inventories with more complex EMC values.

src/main/java/moze_intel/projecte/integration/crafttweaker/mappers/CrTCustomEMCMapper.java
pupnewfster: But wouldn't the proper solution in the script be to just apply the specific override after the tag, and not apply the duplicate value before said tag in the first place?

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
pupnewfster: But none of the call paths pass an empty stack, in fact all of them explicitly ensure the stack is not empty. Which means this will never throw and is pointless clutter instead of just directly calling `NSSItem#createItem`

src/main/java/moze_intel/projecte/emc/mappers/recipe/BaseRecipeTypeMapper.java
pupnewfster: The comments you changed are kind of pointless to change as they are just wording changes; but you still didn't add back to the spot I was asking you to.

